PR #2, opened 2019-03-09
Vorbis dependency changed to faiface/beep

shinlucho:
Vorbis music files are already working, but compared with mp3 and flac it has a different aproach at code level. I've changed it to use faiface/beep so all music formats can be handled the same way.

No idea what Normalizer does tho.


Windblade-GR01:
Thanks for the Commit the normalizer is well:
https://en.wikipedia.org/wiki/Audio_normalization
In reality the normalizer caused problem with the audio (Audio Ducking and cuts in the sound) so it's better removed.
https://en.wikipedia.org/wiki/Ducking

PR #179, opened 2020-09-20
[WIP] character selection in the option menu

marius851000:
Hello. I though the ability to add character via the GUI would be nice (just by drag and drop the good folder in the chars folder). I actually added (and tested) the ability to enable/disable a character from the select.def. I now plan additional things:

- [ ] when the player start Ikemen, scan for newly added character folder, and ask him if he want to enable them
- [x] in the character option menu, display characters not in select.def
  - [x] make the "never loaded" marker not selectable -- removed it
  - [x] create a go function for ls
  - [x] automatically add .def file (for when the folder name isn't the original)
    - [ ] do not do this if the folder isn't renamed ?
    - [x] take in count multiple variation of the same fighter
  - [x] dynamically load the character file when needed (not slowing down start time), and reload them as needed too
- [x] do not need to reload the game after changing a character
- [x] allow to customize the other content of a character in the select.def
- [x] maybe display them in a grid. We can select a character, move them in the grid, and free him. We can press a button to enable/disable it, and another button to edit it's preference. We can also clone a caracter.
  - [x] add the grid display
    - [x] ensure it stick to the left, and try to enable scrolling only when not enought place
  - [x] ability to edit character data
- [ ] ensure to update commit skin change to include this into the option menu
- [ ] add the ability to edit characterselect
- (in another PR) add the possibility to manage stage the same way

I want to if you have any comment to say about this project. (note: its my first go project (but finally learning it wasn't usefull here), and also the first one where I did lua modding in M.U.G.E.N)


marius851000:
Here is what I wish it to look like
dessin

marius851000:
Screenshot_20200924_125832
Here is what it look like now.

marius851000:
This commit change some option related section. I basically added a function dedicated to number selection, and used it were it was applicable in option.

Windblade-GR01:
Marius as you may see, K4thos has released a update that changes a lot of things.
Still you can continue developing the feature in revision the that you have.

I will do the necessary changes to make it compatible with the new version when merging.

marius851000:
Thanks. I should really finish that (when there isn't a lot of important things to do).

SibilantScream:
Arranging select screen grid is still a nightmare even how much facile the new offset parameters feature is.
Also you can't jump from two different slots to one char slot, & rearranging a slot entails recalibrating alot of things & trial & error in most cases.
This is mostly for how outdated Mugen's rigidity, ditching the old format completely is ineluctable for a more flexible one.

marius851000:
Finally finished this message box. It is reusable (haven't reused here, but should nicely replace some message). Here is a screenshot of what it look like now.
image

marius851000:
Also, while doing this, I figured that whenever I write a font, it is drawn a font pixel lower than the coordinate that I indicate. I take it into account when placing them.

marius851000:
I don't plan to return working on this soon. Maybe some day. (it was fun to do).

K4thos:
@marius851000, thanks for spending time working on this feature. Notice that the recent PR closure was unintended. But now that you've stated that it's on hiatus and the fact that it has not been updated for the last 2 years, closing it for the time being seems right. Feel free to re-open it once PR is up to date with current codebase.

PR #444, opened 2022-02-20
Do not register keyboard state for glfw.KeyUnknown

samhocevar:
Setting `sys.keyState[glfw.KeyUnknown]` to `true` can have undesired side effects. For instance, if a joystick button is unmapped (mapped to -1) and multimedia keys are used to change the system volume, `JoystickState()` may return `true`, triggering an event for a button that was explicitly disabled in the config.


Windblade-GR01:
Merged. Thanks for helping in the engine development.

PR #441, opened 2022-02-14
Support for Mugen 1.1 24bpp and 32bpp sprites in SFF files

samhocevar:

PR #450, opened 2022-03-06
Support non-RLE SFF contents

samhocevar:
Hi. This is a reworked version of #441, split into three commits, with the regression fixed, as well as another overlook of my part (`shareCopy()` needs to propagate `coldepth` instead of `rle`).


Windblade-GR01:
Understood, I'll check for any regressions.
If I do not find any I'll merge it latter in the afternoon.

PR #453, opened 2022-03-07
Avoid a string replace call in the text printing code

samhocevar:
This is a more efficient way to print strings that may contain `%`.


ssvb:
This pull request caused https://github.com/ikemen-engine/Ikemen-GO/issues/635

K4thos:
already fixed on github build

PR #502, opened 2022-05-02
Fix issue with some shadows not being rendered

samhocevar:
Shadows of 32-bit sprites were randomly missing because of a typo in the render code.


Windblade-GR01:
Closed because #500 completely removes the part of the code that this commit changes.

PR #531, opened 2022-05-21
Translate *NormalizerLR variable names

SuperFromND:
Translated the names by using the online Japanese dictionary [Jisho](https://jisho.org/) and some guesswork. 
May not be 100% accurate in regards to the terminology around audio compressors/normalizers, but it's better than not being able to read it at all.


samhocevar:
Looks good to me. Thanks!

PR #533, opened 2022-05-22
Fix font rendering problem

Lazin3ss:
This commit fixes a problem introduced in commit: 303734322d24e116dad04eaa21c98da1c4955764, where in-game fonts color wasn't being rendered properly.

Before this commit: https://files.catbox.moe/fchloi.png
After this commit: https://files.catbox.moe/8wv1gl.png

### DIFF
--- a/src/font.go
+++ b/src/font.go
@@ -521,6 +521,7 @@ func (f *Fnt) DrawText(txt string, x, y, xscl, yscl float32, bank, align int32,
 		pal = palfx.getFxPal(f.palettes[bank][:], false)
 	}
 
+	f.paltexs[[4]int32{bank, palfx.eMul[0], palfx.eMul[1], palfx.eMul[2]}] = nil
 	for _, c := range txt {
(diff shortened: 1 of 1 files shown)


K4thos:
this is not a correct fix. If the linked commit was problematic than it should be reverted, not workaround. Upload test case with this font for testing, so that we can see why this particular font behaves like this in the first place.

Lazin3ss:
I understand. Here's the test case, just execute _run.bat_ to see the problem in action. Fonts affected can be found at `data/MKT/font`.

https://files.catbox.moe/12hr23.7z

K4thos:
the problem was caused by this code in your common1.cns, executed at the first frame of Statedef 5900
```lua
[State 0, blank]
type = allpalfx
trigger1 = time = 0
add = -255,-255,-255
time = 3+21*(var(57) != 0)
```

allpalfx affected font pal and that's the version that was stored for the rest of the match (due to optimization introduced in referenced commit). I've reverted the additional optimization, leaving only optimiaztion changes introduced in this commit: 68ec126

PR #559, opened 2022-06-11
New stage property roundpos for rounding position coordinates

GaryCXJk:
A new property has been added to stages. These properties are added on both `BGDef` as well as `BG`, and affect the delta movement of sprites.

`roundpos` defines whether or not a position coordinate has to be rounded before being used by the background sprite's x and y coordinate. You can define a rounding type (`none`, `round`, `floor` and `ceil`). Optionally, you can provide two extra values, which determines the rounding precision. They default to 1. If only the x-value is changed, the y-value will use said x-value.

For example:

```
roundpos = floor, 2, 0.5
```

This says that each sprite's position alignment is rounded down, that the x-value is rounded down to multiples of 2, and that the y-value is rounded down to multiples of 0.5.

You may also use `roundpos.type`, `roundpos.x` and `roundpos.y` if you so choose.

**When would you need this?**

There is a certain edge case, where you absolutely need to have pixel perfect movement. This edge case actually only ever exists in zooming stages, ironically, as there seems to be no issue on regular stages. However, you generally don't want this for every stage, as it will add choppy movement.

This issue actually happens on [one of my own stages](https://area91.garycxjk.com/mugen/stages/nintendo/kirby/kabutterbuilding), where the sprite position has to be exact to the pixel, and again, only in the zoom version.

However, this behavior might actually be beneficial for certain stages that want to more emulate pixelly delta movements.


K4thos:
The problem that I have with this PR is that it seems to be a workaround for differences in how ikemen and mugen 1.1 handles zooming and deltas.

Mugen 1.1 is able to maintain not choppy movement and display zoom version of your stage without extra parameters. So imo the goal should be finding a way to make it work like in mugen 1.1 (rewriting stage.go as needed) instead of adding extra parameters to work around differences.

NeatUnsou:
I believe that the pixel movement of MUGEN1.1 stage changes depending on the options.
If you set mugenversion to 1.1 in [Info], it will be highly accurate, and if not, the position coordinates will be rounded by itself.
Therefore, if mugenversion=1.1 in the Butter Building stage in the example, the screen will be the same as the current IKEMEN GO.

I don't think it is necessary to make this exactly the same as MUGEN 1.1.
For most stages, it would look better if it worked with high precision without having to specify the mugenversion each time.
Still, it is not a bad idea to have a workaround for problematic stages.

K4thos:
if NeatUnsou is right than we could use this implementation to achieve mugen behaviour (especially if the lack of smoothness that comes with it also happens in 1.1). Will post results of my tests here later on.

GaryCXJk:
I do agree we might need to look for a better implementation that emulates MUGEN, though I also do see some value in making it a bit more dynamic for those who want to do more crazy things with with MUGEN by manipulating how the position is rounded.

However, I've also been thinking for a while. I might also have to agree here in that changing the structs might be an added difficulty in the future, provided if it'll be decided that future versions allow for compiled DEF files, as too many changes in the structs might make it more difficult to maintain such files. I know that this might all be future talk, but it's probably important to already keep these things in mind.

ValorZard:
I’d like to chime in and say that adding rounding will cause desyncs online Vegas’s rounding floats is a non deterministic function

K4thos:
thanks for testing. I've made some tests too and the outcome of flooring with precision 1 seems comparable to mugen 1.1 with mugenversion not equal 1.1. I see that you've added a check for mugenversion in latest commit to change default value so it seems good.

I have some suggestions for the implementation tho after checking the code:
1. I don't see a purpose of separating this feature into its own file. Consider moving props.go code to stage.go (for example right above backGround struct)
2. Additional StageProps struct seems not needed (Coord2DProp is enough to implement it)
3. Coord2DProp name is not self explanatory, imo. How about roundPos or posRounding?
4. To match mugen all that is needed is no rounding and flooring. Is there actual use case for normal rounding and ceiling on top of those other 2 methods? If no (or if its super obscure) than having `roundpos.type` accepting string arguments seems excessive.
5. Not sure if having X, Y rounding precision other than 1 is all that useful either.

Maybe a simple boolean int flag (flooring yes or no) for BGDef and BG would be better here?

GaryCXJk:
I originally actually gave it a more specific name, though I decided to make it more general, in case the code needs to be reused for any potential future features, like character modifiers and the likes. Coord2DProp actually is a self-explanatory name, as it defines a property which is a 2D coordinate. This is also why I placed it in StageProps.

StageProps is actually a struct that I added for the simple reason that Golang doesn't support optional parameters. The StageProps object can then be passed through to the backGround objects, so that any prop that's set on a global scale can be used on these structs without having to pass through the entire object.

Most can actually be simplified, though, as emulating MUGEN 1.1 functionality is more of a matter of saying, yeah, it needs to be rounded to the whole pixel. I'll try to simplify it later when I get back to it.

> I’d like to chime in and say that adding rounding will cause desyncs online Because rounding floats is a non deterministic function

In what way? The rounding only gets done after all calculations have been done. Any modifications done at this point won't be saved anywhere, so even if the values don't match up on each machine, it wouldn't matter, because the raw data should be synchronized.

Issue #538, opened 2022-05-26
Feature suggestion for selection screen


Rurouni-X:
**- Skip slot**

A great new feature would be to be able to skip a cell/slot. Currently we can do this by adding a hyphen (some public screenpacks end up using the "do not modify" line), but this ends up requiring a specific select for certain screenpacks. Often the "select.def" ends up being very messy, and this new feature would move the work from "select" to system.def leaving everything cleaner and more organized. The syntax in action:

`cell.<col>.<row>.skip = (0: default), (1: skips*)`

*if "showemptyboxes = 1", the cell doesn't show up;
"moveoveremptyboxes = 1", it's skipped anyway.

An example of these features using the AX2 screenpack.

![preview](https://user-images.githubusercontent.com/68881657/170518046-f66d489f-ca1e-4bd5-9ade-adcc4b8d2e3a.png)

GaryCXJk:
Okay, so, here's the issue. These suggestions can be implemented, however, they won't appear for every screenpack, or at least those that modify the Lua scripts, since the features outlined actually can be implemented through these scripts.

PR #560, opened 2022-06-12
Character select Skip Slot added

GaryCXJk:
As outlined by #538 .

### DIFF
--- a/external/script/main.lua
+++ b/external/script/main.lua
@@ -2007,6 +2007,9 @@ local section = 0
 local row = 0
 local slot = false
 local content = main.f_fileRead(motif.files.select)
+local csRow = 0
+local csCol = 0
+local csCell = 0
 content = content:gsub('([^\r\n;]*)%s*;[^\r\n]*', '%1')
 content = content:gsub('\n%s*\n', '\n')
 for line in content:gmatch('[^\r\n]+') do
@@ -2037,6 +2040,14 @@ for line in content:gmatch('[^\r\n]+') do
 	elseif lineCase:match('^%s*%[%w+%]$') then
 		section = -1
 	elseif section == 1 then --[Characters]
+		csRow = (csCell % motif.select_info.columns) + 1
+		csCol = csCell - ((csRow - 1) * motif.select_info.columns) + 1
+		while not slot and motif.select_info['cell_' .. csCol .. '_' .. csRow .. '_skip'] == 1 do
+			main.f_addChar('-', true, true, false)
+			csCell = csCell + 1
+			csRow = (csCell % motif.select_info.columns) + 1
+			csCol = csCell - ((csCol - 1) * motif.select_info.columns) + 1
+		end
 		if lineCase:match(',%s*exclude%s*=%s*1') then --character should be added after all slots are filled
 			table.insert(t_addExluded, line)
 		elseif lineCase:match('^%s*slot%s*=%s*{%s*$') then --start of the 'multiple chars in one slot' assignment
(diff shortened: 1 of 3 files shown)


K4thos:
to be honest I don't like this idea at all, I don't think it should be merged. You can already title your cell exactly the same as in your change (cell.colNumber.rowNumber.skip) and it will skip just fine (ikemen, just like mugen, skips over non existing char references).

Lua scripts are often spaghetti coded and extremely bloated (desperately need refactoring down the road). Stuff like this is better suited for external modules, imo: https://github.com/ikemen-engine/Ikemen-GO/wiki/Miscellaneous-info#external-modules

I suggest joining discord to gather feedback about features ideas: https://discord.gg/tyASDfmn

Rurouni-X:
> to be honest, I don't think this PR should be merged. You can already title your cell exactly the same as in your change (cell.colNumber.rowNumber.skip) and it will skip just fine (ikemen, just like mugen, skips over non existing char references).

Currently, cells have extra functions like offset, facing and their own cursors.
Considering all this, a method for skipping them doesn't seem like something out of line with the new features.
It's true that we can use "do not modify" lines for this, but with all the improvements wouldn't it be better for the screenpack to handle the select screen design instead of continuing to use MUGEN's hacky method?
Also, with this new feature it would not be mandatory to use a unique/custom select.def like this:
[select.zip](https://github.com/ikemen-engine/Ikemen-GO/files/8885867/select.zip)
shot

This would benefit the end user as they wouldn't have to worry about messing up the roster organization or keeping a backup of select.def.

K4thos:
> Considering all this, a method for skipping them doesn't seem like something out of line with the new features.

a method for skipping them already exists

GaryCXJk:
I do have to respectfully disagree with that assessment, as I did think about the implications of changing it on a script level.

By using the current MUGEN method of skipping cells, you'll run into two issues. The first issue is that, because of the `moveoveremptyboxes` setting being set to 0, you have to set all the remaining boxes to something like randomselect in order for them to be visibly present. Sure, it isn't that big of a deal, but it does bring up a second issue.

Your select.def is bound by the layout.

This is an issue with MUGEN at the moment, where you can't just change a screenpack without having to make sure the select.def fits the current screenpack's character select screen. If a screenpack relies on cells being skipped, and you have already have a full select.def, you'll have to manually copy over each entry, making sure you copy the right amount of lines. With the `slot` feature, you'll run into an even bigger problem, as now, the lines copied over don't have to match the amount of lines that need to be replaced.

Now, while it is true that screenpacks can still require you to overwrite the Lua files, it'd more than likely be a lesser issue, as changing screenpacks again with a different set of Lua files won't likely cause problems with the select.def.

A final point that can be made here is, in theory, this change won't affect existing screenpacks. It'd be a whole different thing if it did. Could this be coded better? Most likely, for example, I really don't like how I added skips by just adding dummy characters, as it really is no different than how it has been done right now, and more ideally, you'd probably want a separate def file to define layouts. But using this feature or not won't matter. In the end, this is mostly a convenience change rather than an absolute requirement.

Windblade-GR01:
On this case I agree with Gary.
Having to copy-paste over 1000 times on a file can not be called "not really a workaround".

It's a incredibly tedious process and it creates problems.

K4thos:
The feature of having the same select.def functional with different screenpacks indeed sounds nice, I have not thought about it from this perspective.

Will merge with little changes (no point in searching for file existence when it's known that slot is dummy from the get go, should save some loading time in larger screenpacks).

Thanks for contribution.

PR #576, opened 2022-06-20
Trials Mode Functionality

two4teezee:
This pull request modifies four Go files to enable a universal Trials Mode, while the Trials Mode itself ships in an externalized Lua module. Changes to the engine are detailed below. 

In char.go, functionality has been added to read and parse a "trials.def" file in the character's folder, which needs to be referenced in the character's definition file. A template can be found in the notes section of the Trials Mode module. After it is parsed, this trials information is stored in an array so it can later be retrieved in Lua via the "gettrialsinfo()" function, defined in script.go. 

Additionally, minor changes were made to lifebar.go and system.go to enable the [timer] functionality even when round time is set to infinite. This is useful in Trials Mode but could also be used in other modes where a timer without being restricted by round time. New state controller: freezeLifeBarTimer(bool), which can be used to freeze and unfreeze the timer, and new trigger: getLifeBarTimer(), which can be used to retrieved the timer value. Note that setLifeBarTimer() will continue to work without issue.

### DIFF
--- a/data/training.zss
+++ b/data/training.zss
@@ -25,7 +25,7 @@ if playerId($pid),ctrl = 0 {
 #===============================================================================
 [StateDef -4]
 
-if gameMode != "training" || isHelper || teamSide != 2 {
+if (gameMode != "training" && gameMode != "trials") || isHelper || teamSide != 2 {
 	# Do nothing, not training more or statedef executed by helper or not P2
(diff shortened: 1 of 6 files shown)


two4teezee:
Will push a new PR for trials mode. Approach is completely different now.

PR #584, opened 2022-06-26
Reduce memory usage when reloading characters with large sff

Eiton:
This commit is a workaround for the memory leak(?) issue where the sprite textures are not released immediately when reloading the match...
It significantly reduces memory usage when you need to frequently reload characters with large sff, on windows 10 at least...


samhocevar:
These changes make sense to me, especially when a lot of characters are loaded at the same time.

PR #619, opened 2022-07-24
Fix camera follow priority

Lazin3ss:
In mugen, the camera always follows the highest player available in the Y-axis, and, when there are no characters left to track (meaning all chars have movecamera Y set to 0), the camera rests wherever it has landed until there's a new character to track. This change deprecates use of "lowest", since it serves no purpose for camera tracking other than setting the lowest value possible to highest before finding the real highest value. I also simplified some math operations.

### DIFF
--- a/src/camera.go
+++ b/src/camera.go
@@ -59,12 +59,8 @@ func (c *Camera) Init() {
 	c.halfWidth = float32(sys.gameWidth) / 2
 	c.XMin = c.boundL - c.halfWidth/c.BaseScale()
 	c.XMax = c.boundR + c.halfWidth/c.BaseScale()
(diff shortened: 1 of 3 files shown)


NeatUnsou:
It seems that when a character makes a super jump at the same time, a zoomout occurs, which should not happen.
It may be that the distance to the ground is included in the conditions for zoomout.

K4thos:
@NeatUnsou, please let us know, if you think this PR can be merged as is, or if it needs additional work

NeatUnsou:
#620 
To prevent the zoomout from inadvertently working on superjumps, needed to make "lowest" work again.
At least this should work fine.

PR #620, opened 2022-07-24
画面内のキャラが同時にスーパージャンプした際にズームアウトが暴発する不具合を修正

NeatUnsou:


Lazin3ss:
Wait! I don't think bringing back lowest was the correct solution for the problem. Instead, maybe, what was needed to do was to remove the `Pow(lowest-highest, 2)` addition [here](https://github.com/ikemen-engine/Ikemen-GO/blob/ae6c763012022acbe3de25abf4e339fc607432d8/src/camera.go#L181). I think this replicates correct behavior from mugen regarding zoom in this context (as it seems Y-axis positionation shouldn't influence zoom at all). I propose reverting this pull request, then removing that addition I mentioned (which was `Pow(MinF(highest, 0), 2)` in my commit).

Also, is there any reason why just removing localcoord[1] addition from boundLo calculation [here](https://github.com/ikemen-engine/Ikemen-GO/blob/master/src/camera.go#L63)? Since that breaks correct boundlow setting. If localcoord[1] addition is removed, gameHeight substraction should be removed too.

NeatUnsou:
In my opinion, the positioning of the Y axis affects the zoom.
Try doing an aerial combo with MVC characters, or with KFM, which allows infinite jumps, and move all the characters to a higher position.
If each Y axis is close to each other, MUGEN will not zoom out, but Pow(MinF(highest, 0), 2) will.
Yes, for Pow(lowest-highest, 2) the zoom changes if all chars have movecamera Y set to 0, but MUGEN does not have this change. I understand that this needs to be fixed.

As for boundLo, surely you are right, gameHeight needs to be removed. It must be fixed.
The reason for removing localcoord[1] is that the values are wrong when the screen ratio changes. I noticed this when I tested with a character that could dive under the ground.

Lazin3ss:
The thing is, in MUGEN, if you use Movecamera = 0,1 on P2 and then move P1 around, you'll see you don't zoomout. Also, in Ikemen GO, with these changes, it will zoomout if you really go up (modify Mountainside temple to expand its boundaries, put verticalfollow in 1, then move KFM up). I think this might be proof highest or lowest don't matter when creating the scale mul for the zoom effect.

Also, here are some comparison videos I recorded performing various movements with kfm in Mountainside Temple (with bounds expanded), all of them in 800x508:

**Screenbound 1, Movecamera X = 1 Y = 1 on both characters**
Both chars moving up:
[Mugen 1.1](https://files.catbox.moe/y7eqv2.mp4)
[Ikemen GO with my proposed changes](https://files.catbox.moe/jhddfq.mp4)

Both chars moving down:
[Mugen 1.1](https://files.catbox.moe/exegsb.mp4)
[Ikemen GO with my proposed changes](https://files.catbox.moe/quhtxf.mp4)

P1 moving up + characters performing various superjump tricks:
[Mugen 1.1](https://files.catbox.moe/ufm5lb.mp4)
[Ikemen GO with my proposed changes](https://files.catbox.moe/7og4qp.mp4)

**Screenbound = 0, Movecamera X = 0 Y = 1 on P2**
P1: moving up:
[Mugen 1.1](https://files.catbox.moe/os5e02.mp4)
[Current Ikemen GO](https://files.catbox.moe/4688p5.mp4)
[Ikemen GO with my proposed changes](https://files.catbox.moe/ozpbp4.mp4)

To sum it up, what I propose is:
- Revert https://github.com/ikemen-engine/Ikemen-GO/pull/620/commits/70a87c95c960a4d9186b4a1c28a59af9014e49d1
- Use neither `Pow(MinF(highest, 0), 2)` nor `Pow(lowest-highest, 2)`, instead just let that line be like this: `tmp = MaxF(220/c.Scale, float32(math.Sqrt(float64(Pow(tmp, 2)))))`
- Remove GameHeight from boundLo assignation

NeatUnsou:
Certainly that behavior will be correct for that condition.
The problem is that the test does not take **tensionhigh** and **tensionlow** into account.
This parameter ignores the description of floortension and verticalfollow and gives a  different camera movement to the stage.
It is not used for the Mountainside Temple stage.

Add this description to the stage and it should work similar to Current Ikemen GO.
```
tensionhigh = 105
tensionlow = -47
```

Of course, they are not exactly the same, because the zoom in Ikemen GO is based on the one implemented in Ikemen SSZ. This was created before tensionhigh and tensionlow were implemented in MUGEN.

These problems should eventually be solved by implementing tensionhigh and tensionlow correctly.
The current Ikemen GO is still inadequate to reproduce them.

Lazin3ss:
I see, I never used tensionhigh/tensionlow before so I didn't know it worked this way, my bad. In that case, maybe we should nest the behavior for this condition in case tensionhigh/tensionlow isn't used. What do you think about this set of changes? https://github.com/Lazin3ss/Ikemen-GO/commit/8f5e0571c77be51e2725397eb2ea8c2811fc4e11

(About the "Highest can't be greater than 0 when working with tensionhigh/tensionlow" thing, I made on the basis that, before implementing camera priority fix, highest was never getting a value above zero, but, given I still didn't get into completely understanding tensionhigh and tensionlow, this can be changed)

NeatUnsou:
I think this is a good change for now.
You should have no problem making a pull request now.

I also forgot that I had added tensionhigh and tensionlow to most of the stages I have, so I apologize for the lack of explanation.

Lazin3ss:
Alright, I added an additional comment and made the PR 🙌

PR #629, opened 2022-08-06
Fix PS3 controller detection

ssvb:
Joystick configuration is done by observing transitions from "some button is pressed" to "all buttons are released" state. See function "f_keyCfg" in "external/script/options.lua": https://github.com/ikemen-engine/Ikemen-GO/blob/0.98.2/external/script/options.lua#L1754-L1761

But right now "getJoystickKey" function returns "-11" for an idle PS3 controller with no buttons pressed (because axes 2 and
5 have -1.0 value in this state). The game does not see the "all buttons are released" state and this breaks the joystick configuration process.

This patch fixes the problem by just skipping axes 2 and 5. So that the "getJoystickKey" function returns "" in an idle state
as expected.

Only tested in Linux. I'm not sure what is the PS3 controller support status in Windows.

### DIFF
--- a/src/script.go
+++ b/src/script.go
@@ -1315,6 +1315,17 @@ func systemScriptInit(l *lua.LState) {
 						} else if axes[i] < -0.5 && i < 4 {
 							s = strconv.Itoa(-i*2 - 1)
 						}
+					} else if joystick[joy].GetGamepadName() == "PS3 Controller" {
+						if (len(axes) == 8 && i != 3 && i != 4 && i != 6 && i != 7) ||
+							(len(axes) == 6 && i != 2 && i != 5) {
+							// 8 axes in Windows (need to skip 3, 4, 6, 7) and
+							// 6 axes in Linux (need to skip 2 and 5)
+							if axes[i] < -0.2 {
+								s = strconv.Itoa(-i*2 - 1)
+							} else if axes[i] > 0.2 {
+								s = strconv.Itoa(-i*2 - 2)
+							}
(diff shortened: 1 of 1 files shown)


ssvb:
Also installed Windows 10 and tested my PS3 controller there. The drivers situation in Windows is quite messy and there are multiple incompatible drivers available: https://www.pcgamingwiki.com/wiki/Controller:DualShock_3 (some of them are emulating Xbox 360 or something else). For example, the reWASD driver from the top of that list looks like this:
win10_rewasd
The `joystick[joy].GetGamepadName()` calls return "" (an empty string), but this driver works fine with Ikemen-GO.

When testing the official Windows driver signed by Sony (which is also used by Steam?), the `joystick[joy].GetGamepadName()` calls return the "PS3 Controller" string just like in Linux. But the buttons/axes layout is different:
win10_official_driver
This joystick configuration doesn't work in Ikemen-GO with this driver for the same reason why it doesn't work in Linux. The sliders for axes 3, 4, 6, 7 are in the rightmost position, and this is interpreted as if some buttons are always pressed.

Here's Linux for comparison:
ps3_idle_state

The screenshots are from the [joysticks](https://github.com/glfw/glfw/blob/master/tests/joysticks.c) test program from https://github.com/glfw/glfw

I have updated the pull request to support both Linux and Windows by skipping the problematic axes. Though now I wonder if it may be even better to just ignore all axes? The PS3 controller has enough normal buttons even without any axes.

PR #630, opened 2022-08-06
Add Explod clip window

Eiton:
This commit adds a new Explod parameter clipwindow. 
```
clipwindow = x1, y1, x2, y2 (float)
    This parameter takes four numbers in the format of clsn which forms a rectangle 
    where pixels beyond the rectangle are not rendered. 
```
This makes things a lot easier when making gauges and animations like the gif below.
goat

### DIFF
--- a/src/anim.go
+++ b/src/anim.go
@@ -725,7 +725,7 @@ func (a *Animation) Draw(window *[4]int32, x, y, xcs, ycs, xs, xbs, ys,
 	}
 	RenderSprite(rp)
 }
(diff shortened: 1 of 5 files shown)


K4thos:
seems useful although I'm not sure about that clipwindow. The same functionality exists for lifebar, screenpack, stage backgrounds, so why not using the same name here? Just window.

Eiton:
Didn't notice it exist in other places. Renamed it to `window`.

PR #660, opened 2022-08-31
Remove calls to gl.ObjectLabel()

samhocevar:
This function only appeared in OpenGL 4.3, and some drivers appear to still provide only OpenGL 4.2. Since its main purpose is for debugging, we should just remove it until there is a way to query for the exact OpenGL version (will require a patch to the Go GL library we use).


lokpro:
let me test

PR #706, opened 2022-11-07
Added prandom trigger

Alceris:
Can't build to test if this works, but I see no reason why it shouldn't.

### DIFF
--- a/src/bytecode.go
+++ b/src/bytecode.go
@@ -488,6 +488,7 @@ const (
 	OC_ex_physics
 	OC_ex_playerno
 	OC_ex_rand
+	OC_ex_prandom
 	OC_ex_ratiolevel
 	OC_ex_receiveddamage
 	OC_ex_receivedhits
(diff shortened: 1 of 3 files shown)


Lazin3ss:
Did a quick test, can't build from it because of an overflow in constant declaration

From the compiler:
`src\bytecode.go:220:2: cannot use iota + 110 (untyped int constant 256) as OpCode value in constant declaration (overflows)`

Alceris:
Yeah, fuck this language.

Lazin3ss:
The implementation is fine in itself, the only thing is, you have to do it via `OC_ex`. Example:
https://github.com/ikemen-engine/Ikemen-GO/commit/52f768a506e26488ae74a76e172b67831abe7bb0

I'd recommend you force-pushing the same commit in your branch, but oriented towards `OC_ex` instead.

Alceris:
I do not understand, but have done.

Lazin3ss:
Engine builds and trigger works as expected, this can be merged.

K4thos:
I don't see a purpose of this. What is this trigger for?

Alceris:
It returns the same value as the previous random trigger without changing it, so you don't need to store the value in order to check the same random number multiple times in one frame.

K4thos:
as discussed on discord this feature is not that useful from zss perspective, so I'm closing it. But thanks for PR.

PR #712, opened 2022-11-11
・BGpalfxがポーズ中に停止しないように修正

NeatUnsou:
・BGpalfxがポーズ中に停止しないように修正
・BGpalfxが1F遅れていたのを修正
・helperのsize.proj.doscaleがintで読み込まれてなかったのを修正


Lazin3ss:
Seems good to me, I'm gonna work over this to finish the palfx BGCtrl implementation via overriding it with bgpalfx when used, then leave an issue detailing how bgpalfx should really be synthesized in stage palfx (since I'm not able to completely solve it myself for now, mostly an issue when using invertall from bgpalfx and how it affects background elements using palfx BGCtrl).

Issue #566, opened 2022-06-15
GameWidth and GameHeight / MajorVersion


potsmugen:
In Mugen 1.1, GameWidth and GameHeight triggers change according to the camera zoom. I just found out that they don't in Ikemen Go.

![gw1](https://user-images.githubusercontent.com/107247004/173867288-1c62163e-d2a2-4e59-858f-6796dffdc9ec.png)

![gw2](https://user-images.githubusercontent.com/107247004/173867751-f96ffac1-22fa-4c48-9291-fa439507e51e.png)

NeatUnsou:
If mugenversion=1.0 is specified in def, GameWidth and GameHeight will be processed the same as ScreenWidth and ScreenHeight.
If you specify a value other than 1.0, GameWidth and GameHeight will be affected by zooming.
The reason for this is that 1.0 characters may use GameWidth and GameHeight to display gauge bars, etc, and there was a problem with them scaling unintentionally  in 1.1.
This specification is intended to prevent that problem, but I understand that it is different from the actual 1.1 specification.

potsmugen:
That actually makes perfect sense. I didn't find it in the documentation.

It's funny because when I work on my characters for Mugen 1.0, I mostly use those triggers so that some things also work correctly in Mugen 1.1, so I'm expecting them to change with zoom. But I think your way makes sense.

Since we touched the subject, I'll take this opportunity to make a related suggestion I had, which is making MajorVersion return 1 if IkemenVersion is specified. I only noticed the GameWidth behaviour because I had to specify a MugenVersion for the characters to work right with the common1 file, even though they are not Mugen versions anymore.

NeatUnsou:
Since MugenVersion and IkemenVersion are treated as different concepts inside IKEMEN GO, it may not be a good idea to treat each with the same trigger.
For example, a character might have IkemenVersion of 1 and MugenVersion of 0 (winmugen).
In this case, it may be confusing for MajorVersion to return 1.
However, a trigger that returns IkemenVersion is necessary.

potsmugen:
Essentially what I'm requesting is that the common1 file works for Ikemen characters the same way it works for MajorVersion characters. Not sure what's the best way to achieve that now that you mention it. I'm not sure even Elecbyte made legacy characters behave differently in those states.

NeatUnsou:
[...]

Issue #658, opened 2022-08-29
[Intermittent] crash on [State -2] when mashing "recovery" command (PP) while in mid air, in chars that use unpatched CvS/POTS system


AxeLanderMoreira:
Steps to reproduce:
While playing single arcade mode (character doesn't seem to matter), if I get hit and try mid-air recovery (that is, press M+H or L+M+H punch buttons simultaneously and repeatedly), the engine will sometimes crash before the recovery is event attempted. Crash invariably occurs due to "panic: runtime error: index out of range" at bytecode.go.

Expected behavior:
It should perform the air recovery move (or none, if not applicable to the selected character), instead of crash.

Platform:
 - Ikemen_GO_Linux compiled from source
I'm specifically using Linux port, but I'm not sure if the issue happens in other platforms. Ikemen-GO Linux compiled from source. I just made sure that the issue still occurs in current HEAD version:

commit 6d0b8391ee0b524f9ade20c254fd2e10022e3ca5 (HEAD -> master, origin/master, origin/HEAD)
Author: K4thos <piotrpit.uek@gmail.com>
Date:   Sun Aug 28 18:18:57 2022 +0200

    Fix sound panning localcoord calc

But it's been happening for quite some time.

Additional context:
I will attach screenshots and stack trace logs.

AxeLanderMoreira:
![Screenshot from 2022-08-29 18-18-48](https://user-images.githubusercontent.com/96396426/187317339-a3bee60b-21d2-4a63-a42e-e3bc3207cbae.png)
![Screenshot from 2022-08-29 20-27-40](https://user-images.githubusercontent.com/96396426/187317341-e8cf7b6f-8731-4290-a909-4bf5c758f53c.png)

Hereis also a log of one occurrence:

2022/08/29 18:16:17 Using OpenGL 4.6 (Compatibility Profile) Mesa 21.2.6 (Mesa Intel(R) HD Graphics 620 (KBL GT2))
Loading module: external/mods/multipleface2.lua
Loading module: external/mods/stagetext.lua
panic: runtime error: index out of range [451] with length 152
goroutine 1 [running, locked to thread]:
github.com/yuin/gopher-lua.(*LState).PCall.func1()
	github.com/yuin/gopher-lua@v0.0.0-20220504180219-658193537a64/state.go:1988 +0x21f
panic({0x8b8260, 0xc002e07ef0})
	runtime/panic.go:884 +0x212
[...]

PR #729, opened 2022-11-23
起き上がりステートのtimeが不正確になっていたのを修正

NeatUnsou:
・起き上がりステートのtimeが不正確になっていたのを修正
・GameHeightとGameWidthの処理がステート奪取側基準になっていなかったのを修正 #566


Lazin3ss:
there's a problem with moving time update to actionFinish, in that some characters break bad when updating it after all characers have been processed. Take a look at the test case from #658, just run a 4vs4 tag with them, then enable AI on all characters and you'll see one of them will get bugged after a while

NeatUnsou:
I didn't think much about it.
This should be fine.

Lazin3ss:
Nice, indeed everything works fine!

PR #750, opened 2022-12-01
New trigger reversalattr

Eiton:
This commit adds a new trigger `reversalattr` which is similar to `hitdefattr`. It checks the attribute parameter of the player's currently-active ReversalDef.


potsmugen:
If I can offer a bit of feedback, I think it'd be more consistent if called "reversaldefattr", even if it's a bit longer.

Eiton:
Yeah. You are right.

PR #769, opened 2022-12-07
Bugfixes: Text SCTRL default color values, AttachedChar font loading

Lazin3ss:
AttachedChars now can render text with their own fonts, since font selection in `Text` SCTRL is now done at a new CharGlobalInfo field for fonts.

Also, a bug regarding `color` parameter from `Text` SCTRL was addressed, where color was not getting set to default values properly, leading to Sprite-based fonts appearing invisible.


Lazin3ss:
Now that I think about it, it wasn't necessary to do the changes to color parameter handling in Text SCTRL, I'll re-do those changes since I feel it's better that way.

Issue #805, opened 2023-01-09
Bug: Engine Crash


djonskin:
**Describe the bug**
I am crashing on startup.

external/script/main.lua:2147: attempt to compare string with number
stack traceback:
external/script/main.lua:2147: in main chunk
[G]: ?

All I could really gather is that something's up with the way the syntax for rmax and rmin is written. After some troubleshooting, a (now closed) PR I did got me back into the program.

**To Reproduce**
I was just trying to START THIS. I'm using the STOCK Elecbyte screenpack, but nothing else should be erroneous. This is with the latest compiled build.
 Here's the steps anyone with a Windows 10 x64 machine can do to reproduce my crash, though I've had no luck so far with another reproduction from another individual.
1: Fresh bin folder (from a new compilation of the latest IKEMEN) with no modifications. This folder has only Data, External and the latest Ikemen GO exe compiled from github.
2: Then add from the Elecbyte pack repository the necessary System.def components.
3: Add fonts in from the pack and it errors on startup.

**Expected behavior**
A normal startup.

**Screenshots / Video**
![image (2)](https://user-images.githubusercontent.com/106702560/211409973-b611ba28-e9da-4a48-8b68-267d58b976f1.png)

**Platform (please complete the following information):**
 - Engine Version -- LATEST IKEMEN BUILD.
 - OS: Windows 10

K4thos:
can't reproduce

djonskin:
I mentioned it's the latest build. Can you link yours then?

djonskin:
Submitted PR for proposed syntax change in main.lua ratiomatches section.

K4thos:
please upload whole build that crashes (drag and drop zip here), without your changes to lua files. Still can't reproduce it based on your #816 steps.

JJMP4:
Hey any updates? I had a similar issue but mine was had the numbers 2155 instead, I believe it's something to do with ratiomatches.

K4thos:
No. I've said in my previous post what has to be provided for us to be able to reproduce it locally.

djonskin:
Now included in description K4thos.

slzero-X:
He followed exactly as you told him, and still not the same results. I detect a "hardware issue"

djonskin:
> He followed exactly as you told him, and still not the same results. I detect a "hardware issue"

[...]

PR #812, opened 2023-01-14
RatioMatches extra code

djonskin:
I had a startup crash that was seemingly hard to replicate for the devs, so Main.lua needed this for the syntax.


K4thos:
please provide a way to reproduce it in this ticket: #805 Change will be made once I can reproduce the problem locally (most likely differently than in this PR, so I'm closing it)

PR #816, opened 2023-01-17
RatioMatches Extra Code

djonskin:
I am still crashing on startup without altering this syntax to how it is. Here's the steps any layperson, I.E. me, can perform to reproduce my crash. 1: Fresh bin folder (from a new compilation of the latest IKEMEN) with no modifications. This folder has only Data, External and the latest Ikemen GO exe compiled from github. 2: Then add from the Elecbyte pack repository, the necessary System.def components. 3: Add fonts in from the pack and get my error. Voila!

external/script/main.lua:2147: attempt to compare string with number
stack traceback:
    external/script/main.lua:2147: in main chunk
    [G]: ?

Speaking with Lazin3ss on it, all I could really gather is that something's up with the way the syntax for rmax and rmin is written. After some troubleshooting, this got me back into the program.

### DIFF
--- a/external/script/main.lua
+++ b/external/script/main.lua
@@ -2144,7 +2144,9 @@ for line in content:gmatch('[^\r\n]+') do
 				rmin = tonumber(rmin)
 				rmax = tonumber(rmax) or rmin
 				order = tonumber(order)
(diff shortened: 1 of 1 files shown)


K4thos:
Changes are never merged blindly, when something can't be reproduced locally. And 8 repetitions of tonumber is not how it will be solved, even if there is a problem there, so this PR won't be accepted. I've reopened your ticket, please continue there.

PR #817, opened 2023-01-18
Add support for "CommonFx" files

Lazin3ss:
This PR introduces a new feature called "CommonFx". Just like most of the other common files, CommonFx is a file array located in `config.json` for general resource  files that can be accessed by all characters in a match. In this case, CommonFx expects definition (.def) files containing information for a "FightFx" package, including .sff, .air and .snd files. A prefix and anim scale can be defined in an info section.

A package .def looks something like this:
```ini
; Example for a Mortal Kombat effects package 
[Info]
prefix = MK ; Obligatory
fx.scale = 1

[Files]
air = mk.air
sff = mk.sff
snd = mk.snd
```
This package can then be accessed by adding/changing the prefix in valid parameters (usually anim/sound ones) from SCTRLs in a character's code:
```coffeescript
playSnd{value: MK5,0}
explod{anim: MK40}
```
This way, characters can access new effects without having to override FightFx from lifebar.

The only obligatory parameter is "Prefix" from [Info] section. This parameter must be any string except for "F" and "S", since these are reserved for the system to call lifebar FightFx and own char files, respectively.

If two packages share the same prefix, the newest package loaded will override the previous one, including all its resource files, no matter if they're empty or not.

### DIFF
--- a/external/script/main.lua
+++ b/external/script/main.lua
@@ -1332,6 +1332,7 @@ function main.f_commandLine()
 	local roundTime = config.RoundTime
 	if main.flags['-loadmotif'] == nil then
 		loadLifebar(main.lifebarDef)
+		loadCommonFx()
 	end
 	setLifebarElements({guardbar = config.BarGuard, stunbar = config.BarStun, redlifebar = config.BarRedLife})
 	local frames = framespercount()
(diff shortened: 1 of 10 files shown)


Lazin3ss:
Limitations we should think about:
- Right now, "prefix" accepts any string, meaning, numeric strings might compile and can mess up with the game. We might have to ensure prefixes only contain A-Z characters.
- "Prefix" accepts a string of unlimited length. Maybe we should limit the character length to avoid ridiculously long prefix names.

Notes:
- CommonFx are loaded using a `loadCommonFx()` lua function that's called just after lifebars are loaded in main.lua.

PR #819, opened 2023-01-19
Updated Common files to make use of CommonFx

RealFoobs:
This pull request creates a new CommonFx definition (GO.def), and changes the existing common files to use its prefix. 

gofx.air, gofx.sff and gofx.snd are functional copies of the default fightfx.air, fightfx.sff and common.snd files, they only exist so previously created hitspark sets for the mugen engine can work with ikemen without locking the user out of the effects included in the common files (dizzy, guard crush...)


K4thos:
please also fix the typos
"Do note all all"
"This prefix is to used"
after that it can be merged, thanks

RealFoobs:
Done and done.

PR #1068, opened 2023-03-18
fix: Fixed guarding animation lag

potsmugen:
Eliminated one frame of animation lag when switching from high to low guard and vice-versa. No effect on gameplay.

### DIFF
--- a/data/common1.cns.zss
+++ b/data/common1.cns.zss
@@ -276,6 +276,7 @@ if command != "holdback" || !inGuardDist {
 # GUARD (start)
 [StateDef 120; type: U; physics: U;]
 
+call HiToLoAndLoToHi();
 if time = 0 {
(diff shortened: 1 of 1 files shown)


K4thos:
sounds like a "fix", not "other". I assume it fixes #1069? I'm not sure it "other" as a category even makes sense since everything that I can think of fits one of the following: 
- build: Changes that affect the build system, external dependencies, CI configuration
- docs: Documentation only changes
- feat: A new feature
- fix: A bug fix
- perf: A code change that improves performance
- refactor: A code change that neither fixes a bug nor adds a feature
- style: Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc)

potsmugen:
I was in doubt about labelling it a fix since it's so minor. It's not related to #1069. I just found both issues almost simultaneously.

PR #1075, opened 2023-03-20
feat: update rect, text, color, and hooks 

Wintermourn:
This PR contains changes that bring my code up-to-date with my current standards and skills, adding in some new features and hopefully (slightly) less memory usage.

`rect`, `color`, `text`, and `hook` have been separated into their own files, all loaded together through `external.script.lua-tools.init`.
`rect`, `color`, and `text` now have their own shared metatables instead of making a new one for each instance.

- default values are automatically used.
- `text:create` and `text:new` are now `text.create` and `text.new`.
- `color:new`, `color:fromHex` are now `color.new` and `color.fromHex`.
- `rect:create` is now `rect.create`.

`rect:update(t)` now checks the key `color`, which will update it's `r`, `g`, `b`, `src`, and `dst` values.
`color:toHex` is shorter, using `string.format`.

New functions for hooks:

- `hook.on(list, callback)`: adds a callback to a hook list without a name.
- `hook.removeOn(list, callback)`: removes a callback from the recurring list, named or not.
- `hook.once(list, callback)`: adds a callback to a hook list which only runs once.
- `hook.removeOnce(list, callback)`: removes a one-time callback. Probably only useful if one mod's trying to cut another off.

One time callbacks are run _before_ recurring ones.

New functions for colors:

- `color.fromHSL(h, s, l, src, dst)`: Creates a color using the HSL format.
- `Color:toHSL()`: Returns `h`, `s`, `l`, `src`, and `dst` values.

Another file has been made to store extra functions that extend existing parts of Lua, currently only including `table.getKey(tbl, value)`.


K4thos:
It looks very good and can be merged once stable version 0.99.0 is published. However, please keep in mind that a major Lua refactoring is currently in progress. Therefore, the changes made and the file structure implemented will not be kept once that branch is merged. For this reason, I am not encouraging people to work with the current scripts and instead wait for the refactoring to be finished.

The general approach is similar to yours but scope of object-oriented changes affects most of the code. Currently, the scripts have been separated into 41 classes and subclasses with inheritance, each in a different file, and more will be added. Here is an example of your old color functions following the style and commenting that the Lua refactored code strictly follows: https://gist.github.com/K4thos/6893e5b458cbfa1358f318988199a242

Wintermourn:
I think I remember hearing something about the refactoring a while ago but I didn't know whether it was being worked on or if it was public. The code and comments look nice, and I look forward to seeing and working with the rest of it. 
I am a bit confused by the code making a global variable and then returning that. Is that intentional?

K4thos:
That last return can be most likely skipped, since it's a global but not using local for this is intentional. Without global either VSC or IntelliJ extension that this type of commenting is aimed at wasn't able to provide all hints, suggest types, code completion or following functions between files, if I remember correctly, but I don't remember the details right now.

K4thos:
I've closed the PR since all its features have been implemented into the previously mentioned refactored scripts (yes, still in progress), so nothing will be lost. I believe there isn't much benefit in merging these changes into the current codebase, only to replace them later with an alternative refactoring approach.

PR #1103, opened 2023-03-29
midi solution for sound.go

Mike77154:
maybe this is the thing that can help for making midi compatible ikemen


K4thos:
In the current form the code can't even be compiled, so I assume was written blindly, without testing, so it's nowhere near ready for PR . Also if you're adding a new dependency please update go.mod and go.sum. Feel free to open new PR after you're sure that the code works as intended.

PR #1111, opened 2023-03-30
feat: Extra training mode guard modes

potsmugen:
Added a couple fairly common features to the training dummy's guard modes.
- Auto mode renamed to All mode
- New Auto mode: dummy blocks automatically for a short time after getting hit
- Random mode: dummy blocks with a 50% chance


K4thos:
your changes to lua files replace whole files content, so I assume you are overwriting new lines to different format or something like that. This makes it impossible for git to correctly show what part of code has been changed and makes the review harder since I have to use external program to compare changes. What program you're using to edit files? It's not usual for programs to change newlines encoding.

potsmugen:
Interesting. I had no idea and just used Notepad++. If you tell me which program to use instead I can redo the changes since they're not complex.

K4thos:
I also use notepad++ and it doesn't happen for me. Not sure what's up.

potsmugen:
Weird. The end of line character is the same as before from what I can tell, and an external program shows the differences properly.

PR #1121, opened 2023-04-01
fix: Simplified state 51

potsmugen:
- This state is used in Mugen to simply make the character fall down to the ground and land. As such it now behaves the same way as it did there and does not alter the character's animation


K4thos:
I see in mugen common1 that it does this change too, so not sure how it helps compatibility. Either way please comment out those lines, not remove them completely, so that it's easy to revert.

PR #1149, opened 2023-04-07
feat: add const(data.fall.defence_up) trigger

potsmugen:
- Like Mugen, one could not read the character's fall.defence_up constant directly, having to reverse the fall.defence_mul formula to get it. It is now accessible


K4thos (changes requested):
Please also add this to script.go "const"
```
		case "data.fall.defence_up":
			ln = lua.LNumber(c.gi().data.fall.defence_up)
```

potsmugen:
Looks like I missed one. Sorry. It is done.

PR #1265, opened 2023-06-03
feat: Explod RemoveOnChangeState parameter

potsmugen:
- Setting this parameter to 1 makes the Explod be removed if the character changes state
- Fix: if a character is KO'd, their get hit animation is now updated to "fall type"


potsmugen:
Did not notice the other active PR. Will merge this one later.

PR #1275, opened 2023-06-09
feat: new Math triggers (Clamp, Sign, Atan2, Rad, Deg, Lerp) and char triggers (Angle, Scale, OffSet, Alpha)

Wftk:
- **fix: Rgba sprites mask inverse when palfx invertall = 1.**
Before fix:
ikemen001
After fix:
ikemen002
[Char link](https://mega.nz/file/Hk5wmYrR#Ohmj1O49TANZbcwdfwoAN6dlPH-EA6ul0um4H78yFYM)

New Math triggers:

- Clamp(x,min,max) (float)
Return a value clamped to an inclusive range of two specified numbers

- Sign(value) (int)
Returns the sign of a real number. If value < 0 return -1. If value 0 return 0. if value > 0 return 1.

- Atan2(x,y) (float)
Returns the arc tangent of the two numbers x and y.

- Rad(value) (float)
Converts degress to radians.

- Deg(value) (float)
Converts radians to degress.

- Lerp(a,b,amount) (float)
Linear interpolation between two values.

New Char triggers:

- Angle (float)
Gets the value of the player's angle.
Example: "trigger1 = angle < 90" 
- Scale (float) (Arguments  x, y) 
Gets the value of the player's scale.
Example: "trigger1 = scale x< 2 || scale y > 2" 
- OffSet (float) (Arguments  x, y) 
Gets the value of the player's offset.
Example: "trigger1 = offset x > 20 || offset y > 15" 
- Alpha (int)  (Arguments  source, dest) 
Gets the value of the player's alpha.
Example: "trigger1 = alpha source > 128 || alpha dest < 64"

### DIFF
--- a/src/bytecode.go
+++ b/src/bytecode.go
@@ -216,13 +216,6 @@ const (
 	OC_helperindex
 	OC_p2
 	OC_stateowner
-	OC_angle
-	OC_scale_x
-	OC_scale_y
-	OC_offset_x
-	OC_offset_y
-	OC_alpha_s
-	OC_alpha_d	
 	OC_rdreset
 	OC_const_
 	OC_st_
@@ -539,6 +532,13 @@ const (
 	OC_ex_envshakevar_time
 	OC_ex_envshakevar_freq
 	OC_ex_envshakevar_ampl
+	OC_ex_angle
+	OC_ex_scale_x
+	OC_ex_scale_y
+	OC_ex_offset_x
+	OC_ex_offset_y
+	OC_ex_alpha_s
+	OC_ex_alpha_d
 )
 const (
 	NumVar     = 60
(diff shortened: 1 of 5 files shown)


K4thos:
some changes before the PR can be merged:

1. These should be changed to OC_ex variants:
OC_angle
OC_scale_x
OC_scale_y
OC_offset_x
OC_offset_y
OC_alpha_s
OC_alpha_d

2. script.go references every single existing trigger, even if adding them is not needed because of exisitng lua variants (in such case they should be added as comments). Following triggers are missing in script.go:
atan2
clamp
deg
lerp
rad
sign

3. In script.go the function is named offsetDest instead of alphaDest.
4. 
5. Please keep the alphabetical order when adding new functions to that file (also applies to other files, if they follow such order, let's keep the existing conventions)

K4thos:
This branch has conflicts that must be resolved

PR #1292, opened 2023-06-14
other: Adding "required" commands to common.cmd

RealFoobs:
Added BB and FF commands to the common.cmd file, as they are listed as required commands in KFM's own cmd. Despite this explicit warning some mugen characters choose to ignore this convention (presumably because they override the run and backdash states).

This might seem extremely minor, but if has a case use: if one wants to implement any function through common files that requires double tapping (like creating a custom state based on 105/106) one can use `command = "FF"` without worrying a stray character will crash the engine.

### DIFF
--- a/src/compiler_functions.go
+++ b/src/compiler_functions.go
@@ -584,8 +584,8 @@ func (c *Compiler) explodSub(is IniSection,
 		explod_id, VT_Int, 1, false); err != nil {
 		return err
 	}
-	// First, we see if postype parameter was declared, to see if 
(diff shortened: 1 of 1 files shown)


potsmugen:
Could use the buffer.time = 1 line like the other commands. The reason it was added to them is so that each character's own default buffer time doesn't apply to the common commands.
The only problem I see with this is that it might give some (probably poorly coded) characters unintended dashing commands.

RealFoobs:
Dully noted. I don't want to introduce any unwanted behavior on exiting characters, I changed the commands to blanks.

PR #1306, opened 2023-06-24
fix: lying down get hit behavior

potsmugen:
A few changes to make it behave more like Mugen:
- The fall defence bonus now stacks every time the character hits the ground
- Lie down recovery time only counts down in state 5110
- Hitting a character that's in state 5101 with a sweep now correctly sends them to state 5070
- The limit on hitting a character off the ground now depends on how much recovery time the char has left rather than how many times they have been bounced
- The value of PrevStateno when the character is hit into a custom state is more accurate to Mugen

Also:
- Dizzy and guard points recovery is no longer affected by defense multipliers


K4thos:
looks like your commits have been squashed before merge, sorry about that

potsmugen:
No problem. I should've squashed two of those anyway.

PR #1344, opened 2023-07-25
fix: Explicitly set GLSL shader versions

SuperFromND:


Lazin3ss:
This approach might get into trouble with this piece of code that adds personalized code for OpenGL ES implementations:
https://github.com/ikemen-engine/Ikemen-GO/blob/master/src/render_gl.go#L82-L84

my recommendation is to force-push a new commit where the version string is added in the context of the lines provided above, probably inside an "else" statement

SuperFromND:
Right, interesting results from me trying that. I tried just bolting on an else statement:
```go
	if strings.Contains(gl.GetString(gl.VERSION), "ES") {
		src = "#version 100\nprecision highp float;\n"
	} else {
		src = "#version 120\n"
        }
```
But that results in the program closing almost immediately:
```
2023/07/25 11:57:21 development
panic: Link error: Vertex info
-----------
(0) : error C5145: must write to gl_Position


goroutine 1 [running, locked to thread]:
main.linkProgram({0xc94a0?}, {0xc0?})
        github.com/ikemen-engine/Ikemen-GO/src/render_gl.go:110 +0x208
main.newShaderProgram({0x7ff6ab5b4759?, 0xc000000001?}, {0x7ff6ab5b4f80, 0x567}, {0xc000040000?, 0xc0000c9550?})
        github.com/ikemen-engine/Ikemen-GO/src/render_gl.go:54 +0x65
main.(*Renderer).Init(0x7ff6ab89ede0)
        github.com/ikemen-engine/Ikemen-GO/src/render_gl.go:208 +0x325
main.(*System).init(0x7ff6ab4e5b40, 0x18?, 0x0?)
        github.com/ikemen-engine/Ikemen-GO/src/system.go:419 +0x505
main.main()
        github.com/ikemen-engine/Ikemen-GO/src/main.go:86 +0x19f
```

At a guess, I then changed that line to:
```go
		src = "#version 150\n"
```
Which *doesn't* crash, but results in no visual output (the rest of the engine runs fine, it just doesn't draw anything):
image
Using `fmt.Print(src)` it seems as if that if-else block *replaces* the shader in question entirely with the header, instead of prepending it like it's probably intended to, which I'm assuming is an unrelated bug.

Modifying the line yet again (and fixing the aeforementioned bug):
```go
	// Might be necessary for the WebGL build
	if strings.Contains(gl.GetString(gl.VERSION), "ES") {
		src = "#version 100\nprecision highp float;\n" + src
	} else {
		src = "#version 120\nprecision highp float;\n" + src
        }
```
Will indeed produce functional shaders as intended.
image

Lazin3ss:
That's true, sorry about that, it was my mistake when uploading the original PR, thanks for fixing.

SuperFromND:
Noted! Wasn't sure if the precision was *necessary* under GLSL 1.2 or not, removing now. 👍

Issue #1334, opened 2023-07-22
Zoom camera borders inconsistency


potsmugen:
### Describe the bug

I have a stage that worked fine in Mugen 1.1 and in the builds from a couple weeks ago, that suddenly started zooming to parts of the floor that should not be visible.

### To Reproduce

1. Try this stage:
[WaterfallDojo.zip](https://github.com/ikemen-engine/Ikemen-GO/files/12136663/WaterfallDojo.zip)

### Expected behavior

Same as Mugen 1.1.

### Screenshots / Video

Mugen 1.1:
![mugen007](https://github.com/ikemen-engine/Ikemen-GO/assets/107247004/7275625f-f0b2-4e2b-bca4-e48dff28b586)

Nightly:
![ikemen008](https://github.com/ikemen-engine/Ikemen-GO/assets/107247004/860f031b-99a3-4acb-a9b4-1cfe71715b2b)

### Engine Version (or source code date)

22/07/2023 (worked in 07/07/2023 build)

### Operating system

Windows

### Extra context or search terms

This might be an issue for another time, but while posting this I noticed that the camera zooms out further in Mugen than it does in Ikemen.

Lazin3ss:
To fix this bug, a separation between zoom scaling and other bg positioning variables (delta, start, etc) should be made. Right now they're too entangled and modifying one will alter the other. For example, setting Delta Y to 0 on floor BG will "fix" the positioning when zooming out, but modifying Delta Y on MUGEN 1.1 has no effect on Y-axis positioning of BG elements after zooming out in a stage.

K4thos:
From what I understand it has been fixed by #1346

Issue #1319, opened 2023-07-12
If playing on widescreen, the camera "skips" when the characters jump. yDelta is also different.


Hexioum:
### Describe the bug

All my stages seem to have this problem where the delta movement is wrong (Probably because IKEMEN now reads a parameter that was being ignored)
But the main issue is the camera movement when characters lift off the ground or are about to land.

This problem does not happen with the stages bundled with the program.

### To Reproduce

1) Set the screen to 1280x720
2) Load any of these 2 stages [2stages.zip](https://github.com/ikemen-engine/Ikemen-GO/files/12023341/2stages.zip)
3) Jump

### Expected behavior

A smooth camera movement, which does work on the stages bundled with the program.

### Screenshots / Video

IKEMEN v0.99:
https://github.com/ikemen-engine/Ikemen-GO/assets/36753605/f3035761-94a7-4906-99fb-b03b71ccc8b9
https://streamable.com/zsnqmp
MUGEN 1.1:
https://github.com/ikemen-engine/Ikemen-GO/assets/36753605/356519ef-9ea9-463f-9c27-83ba059512cd
https://streamable.com/4hkxtx

*Edit: Added a streamable link in case the video upload didn't work.

### Engine Version (or source code date)

v0.99.0-rc.1 and nightly from Mar 9

### Operating system

Windows

### Extra context or search terms

Vertical movement, ydelta. Camera behaviour.

Lazin3ss:
Campera skipping is related to mishandling of drawOffsetY field from `StageCamera` struct in camera.go. Delta issue... I'm not sure, I would need further understanding of background drawing to find a solution.

NeatUnsou:
The issue with ydelta seems to be related to hires=1
currently, ydelta is broken for all stages that are using hires

PR #1346, opened 2023-07-27
fix: regression in zoom camera

NeatUnsou:
partially revert the changes that were causing the issue with #1334 in #1323
the problem in #1319 is still fixed

### DIFF
--- a/src/camera.go
+++ b/src/camera.go
@@ -75,14 +75,14 @@ func (c *Camera) Reset() {
 	c.XMin = c.boundL - c.halfWidth/c.BaseScale()
 	c.XMax = c.boundR + c.halfWidth/c.BaseScale()
 	c.ExtraBoundH = ((1 - c.zoomout) * 100) * (1 / c.zoomout) * 2.1 * (float32(sys.gameHeight) / 240)
-	c.boundH = MinF(0, float32(c.boundhigh-c.localcoord[1])*c.localscl+float32(sys.gameHeight)-c.drawOffsetY)
+	c.boundH = MinF(0, float32(c.boundhigh-c.localcoord[1])*c.localscl+float32(sys.gameHeight)-c.drawOffsetY) - c.ExtraBoundH
 	c.boundLo = MaxF(0, float32(c.boundlow)*c.localscl) //-c.drawOffsetY)
 	//if c.boundlow < 0 {
 	//	c.boundLo += float32(c.boundlow) * c.localscl
 	//}
 	xminscl := float32(sys.gameWidth) / (float32(sys.gameWidth) - c.boundL +
 		c.boundR)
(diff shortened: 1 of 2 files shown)


Lazin3ss:
CameraZoomYBound reversion seems reasonable (I can't remember why I made that change, probably to fix a discrepancy with MUGEN 1.1 I can't remember right now), but I don't think ExtraBoundH change should be reverted. That change was done to emulate a MUGEN 1.1 behavior regarding boundhigh fluctuation based on zoom scaling. In MUGEN 1.1, boundhigh gets wider based on how much the stage was zoomed out.

MUGEN 1.1:
https://github.com/ikemen-engine/Ikemen-GO/assets/23774987/116836e7-3bfc-4984-90d5-8b3d7f8915be

Ikemen GO without zoom-based ExtraBoundH fluctuation:
https://github.com/ikemen-engine/Ikemen-GO/assets/23774987/66973266-5408-4e9a-8aa1-7a502bab0837

Ikemen GO with zoom-based ExtraBoundH fluctuation:
https://github.com/ikemen-engine/Ikemen-GO/assets/23774987/8122804f-48bf-4720-bce5-b79f1a7a64ec

NeatUnsou:
I understand that the changes made to ExtraBoundH are intended to emulate the behavior of MUGEN 1.1
However, this behavior has a limitation that boundhigh is not reached unless the screen is zoomed out sufficiently.

As a consequence of this limitation, there is an issue when performing aerial combos with MVC chars.
Both chars stay close to each other while being airborne, and since the zoomout doesn't occur, the Y-scroll stops midway, preventing it from reaching the intended boundhigh limit.

In my view, this as one of the major flaws in MUGEN 1.1, and I am opposed to replicating it in ikemengo. 
I feel the existing ikemengo behavior where boundhigh remains constant regardless of zoom scaling state makes more sense.

Lazin3ss:
Indeed it does, that being said, there might be MUGEN stages that are built with this limitation in mind, and not replicating it could potentially break them.

I think this might be enough of reason for us to start implementing `ikemenversion` flag on stages. That way, we can start pushing forward stagemaking towards what we think it makes sense (as opposed to complying to elecbyte rules), while, at the same time, mantaining full backwards compatibility with MUGEN 1.1.

NeatUnsou:
reverted the zoom scaling behavior that adjusts ExtraBoundH like MUGEN 1.1, making it work only for stages with low zoomdelta values to avoid breaking the existing behavior. This ensures that stages designed with the assumption of MUGEN 1.1 behavior won't be affected while maintaining the functionality of the current ikemen go.

I believe this change maintains compatibility, but I agree with the suggestion that a new parameter might be necessary in the future, as the camera behavior being altered by zoomdelta seems strange way.

Lazin3ss:
We can leave it like this until the next version, where we can add the ikemenversion flag.

PR #1443, opened 2023-10-13
fix: Trans sctrl during hitpause bug

Wftk:
When trans sctrl active and if ignorehitpause = 0 sctrl in ikemen dont reset during hitpause
Examples 

Ikemen:

https://github.com/ikemen-engine/Ikemen-GO/assets/132805988/063d39ca-0122-4d11-be23-3816698f49d2

Mugen: 

https://github.com/ikemen-engine/Ikemen-GO/assets/132805988/cd2c168f-c061-4043-a358-f0ee0d1e1e75

Fix: rgba sprites palfx add param fix for 1.1 blending if summ of src and dest alpha = 256(e.g 128x128, 96x160 etc)

### DIFF
--- a/src/char.go
+++ b/src/char.go
@@ -5720,6 +5720,10 @@ func (c *Char) actionPrepare() {
 			c.angleScale = [...]float32{1, 1}
 			c.offset = [2]float32{}
 		}
+		//Trans reset during hitpause if ignorehitpause = 0 fix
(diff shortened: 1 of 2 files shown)


potsmugen:
It turns out Ikemen was replicating Winmugen in this case, which makes sense given its history. I'll add a version check.

PR #1483, opened 2023-11-06
fix: duplicating commands

### LINKED ISSUE #1474: Same command can be completed via multiple inputs simultaneously
### Describe the bug

I never ended up posting this one. In a nutshell, if you perform a command successfully, other commands with the same name will not be reset. This creates some problems, especially when the char uses negative edge, because the command will register once on button press then twice on button release.

### To Reproduce

1. [Download Reu's Evil Ken](https://network.mugenguild.com/pots/hosted.html)
2. Perform a hard Shoryuken close to the opponent
3. Most of the time he'll immediately do the followup (without flames)

Can be reproduced with whichever chars check the same command in succession. Another good way to test is in chars with custom combo and negative edge (like Warusaki3's for instance).

### Expected behavior

Completing a command should reset every other command with the same name. I've been trying to code it like that but without success yet (so be my guest if you'd like).

### Screenshots / Video

_No response_

0.99 and as far back as I can remember

### Operating system

Windows

_No response_

potsmugen:
- If a command is completed successfully, the buffer for every command with the same name is now reset
- Fixes #1474

### DIFF
--- a/src/input.go
+++ b/src/input.go
@@ -1050,27 +1050,35 @@ func (ce *cmdElem) IsDirToButton(next cmdElem) bool {
 		return false
 	}
 	// This logic seems more complex in Mugen because of variable input delay
-	for _, k := range next.key {
-		// Yes if second element includes buttons
+	// Not if first element includes button press or release
+	for _, k := range ce.key {
 		if k >= CK_a {
-			return true
+			return false
 		}
 	}
-	for i, k := range ce.key {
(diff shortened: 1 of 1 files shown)


K4thos:
This branch has conflicts that must be resolved

potsmugen:
Dang. Sorry. Last minute changes caused it. I'll redo the PR after I do a few more commits.

PR #1557, opened 2023-12-12
feat: animframe(numclsn) parameter

wily-coyote:
This PR adds two new AnimFrame parameters, `NumClsn1` and `NumClsn2`, which return the amount of Clsn* boxes for the character's current animation frame.

### DIFF
--- a/src/bytecode.go
+++ b/src/bytecode.go
@@ -400,8 +400,6 @@ const (
 	OC_ex_wintime
 	OC_ex_winperfect
 	OC_ex_winspecial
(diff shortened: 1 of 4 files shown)


potsmugen:
These kind of seem like they could be integrated in the AnimFrame trigger.

wily-coyote:
I actually haven't thought of AnimFrame before opening the PR. I'll move these triggers to AnimFrame, since that makes the most sense.

K4thos:
please edit script.go accordingly. Search for animframe in code.

PR #1602, opened 2024-01-25
refactor: file system access where ioutil is used is replaced by os

formula1:
I'm doing this so I can isolate packages that would be replaced by web assembly compatible packages


K4thos:
what's the reasoning behind this change?

PR #1616, opened 2024-02-05
feat: Added QHD resolution option in settings

updeshxp:
I did some playing with 2560x1440 resolution and did not see any issues, Since it was missing in default options, hence added it.


potsmugen:
That seems like an odd resolution to have by default. It will probably look the same as 1920x1080 since _at best_ assets are made for that resolution. I think people who want it can already use the custom resolution option.

PR #1624, opened 2024-02-10
refactor: vertex buffer loading

Eiton:
Add a new vertex buffer for stage 3d model
Improve performance for stages using high poly models

### DIFF
--- a/src/render_gl.go
+++ b/src/render_gl.go
@@ -251,8 +251,9 @@ type Renderer struct {
 	spriteShader *ShaderProgram
 	vertexBuffer uint32
 	// Shader and index data for 3D model rendering
-	modelShader *ShaderProgram
-	indexBuffer uint32
+	modelShader       *ShaderProgram
+	stageVertexBuffer uint32
+	stageIndexBuffer  uint32
 }
 
 //go:embed shaders/sprite.vert.glsl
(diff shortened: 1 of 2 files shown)


Daru00:
is there an specific stuff to make it work.
tested all the 3d stages all of them black screen and several other users pointing to the same issue.
so wondering does this work or not?? specially seeing the issues with the stages.
https://00gamers.wordpress.com/2024/02/15/just-a-ranting-about-ikemen-go/

Eiton:
@Daru00 
The only specific stuff is to make sure you are using nightly build, not 0.99 or older version.
 
There is only 1 dev working on the 3d rendering so it is impossible to make sure it works on every computer ( Not even the top gaming studios can! You can see complaints about crashing and graphic problems on Steam forum of popular games). The best thing I can do is to guess the cause of the problem base on the info provided and give suggestions that worked on other people facing similar problems.

So PLEASE provide as mush detail as possible. Engine version, graphic settings, OS version, GPU and anything you can think of. Black screen is not enough.
Also, please create an issue instead of commenting under the pull request so that it is better tracked.

Daru00:
after checking the code,i noticed it won't work no matter what.
there are lot of issues on coding either on regular stages and 3d stages,as the guy from that site said ikemen go does indeed have an issue with opengl since in the code it requires 1.3 but in paper asks 3.0.
After decompiling the nightly from 2 days ago is clear this isn't intended to work,so as the guy from that site site "is just cheap clickbait" and "is a dream further in the future".
PC 1 : Intel Pentium D 2.8 Ghz,SSD,Saphyre Radeon HD 5400 1GB,windows 7 black screen.
PC 2: Amd quad core 2.6 Ghz,GPU R5 graphics 3 GB,windows 10 black screen.
PC 3 : Intel Evo MSI prestige 16,windows 11 black screen.

the reason of commenting is becuase you are the only one active doing the so called 3d stages so going directly at you is easier then posting something that will get ignored.
Also 3D is simple to make sure it works for everyone (big and small companies do that all the time,and the complains are made by random users that pirated the game or use a PC lower then minimal requirements) and is fixing the issues with opengl (open from 6 years ago) and the issues with stages (again open since 6 years ago)

PR #1639, opened 2024-02-19
refactor: Replace Filesystem Calls - WIP for Buildable for Wasm

formula1:
Replaced most places where the os filsystem with an abstraction that can be replaced depending on compile target (complete wasm compatibility being the eventual goal).

I Ignored
- Windows only - https://github.com/ikemen-engine/Ikemen-GO/blob/develop/src/stdout_windows.go

Note, It successfully built for Linux, not sure about other environments.

### DIFF
--- a/src/bytecode.go
+++ b/src/bytecode.go
@@ -4,7 +4,6 @@ import (
 	"encoding/binary"
 	"encoding/gob"
 	"math"
-	"os"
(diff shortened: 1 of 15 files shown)


formula1:
If you want to track my personal progress https://github.com/formula1/ikemen-go-js-bindings

Issue #1652, opened 2024-02-23
Verticalfollow more jerky / faster in Ikemen than in Mugen 1.1


Kasasagi77:
### Describe the bug

The scrolling when jumping is visibly more jerky, and almost assuredly faster than in Mugen 1.1, for the same value of `verticalfollow`. The difference is easily visible.

Tested on newest `develop` as of making this issue (1518892e).

### To Reproduce

You can use my previously uploaded stage for testing:  [graveyrd.zip](https://github.com/ikemen-engine/Ikemen-GO/files/14324642/graveyrd.zip). This `zip` contain a zooming stage with different/separate `.def` variants, use the Mugen 1.1 variant (`*-mugen.def`).

Jump in Mugen 1.1, jump in current Ikemen, observe the difference.

The stage uses zoom, however it is not clear to me of this issue is zoom-dependent - is seems faster in this stage regardless of the current zoom value, however the stage has zoom range fully in the zoom-out area (zoom < 1.0), so the behavior may yet be different for zoom >= 1.0.

### Expected behavior

The overall speed and the speed curve of vertical scrolling should be the same in Ikemen Go and in Mugen 1.1.

### Engine Version (or source code date)

2024-02-23

### Operating system

Linux

potsmugen:
Yeah. I don't think it's that noticeable, but if you output Camerapos Y to the clipboard you can see that Ikemen is slightly off compared to Mugen. The difference is under 1 pixel though.

It also looks like Mugen has a camera lag while Ikemen does not. You can see it in this stage if you jump with KFM and frame step. When KFM lands, the camera is back in its original position in Ikemen, but in Mugen it takes a few frames to get to that position. This lag is probably why it feels smoother in Mugen. Maybe it's a feature we should port?

Another test case is adding this to KFM's light punch:
```
[State 200, Test]
type = posset
trigger1 = time = 5
y = -50
```
In Ikemen, at time = 5 the camera will teleport to the new position. In Mugen, it takes about 12 frames to get there.

Kasasagi77:
~It may be that it is actually the lag that is causing what I am seeing as "jerky" camera movement?~

[...]

PR #1757, opened 2024-04-26
fix: Target camera height/Y when using verticalfollow and zoom

Kasasagi77:
This is the biggest change/fix for #1652, however please do not close the issue yet - the scrolling still seems to be choppy to me and I am still testing it.

@Eiton - you may want to review as this changes the camera code again.

### DIFF
--- a/src/camera.go
+++ b/src/camera.go
@@ -255,7 +255,7 @@ func (c *Camera) action(x, y, scale float32, pause bool) (newX, newY, newScale f
 				if !c.ytensionenable {
 					newY = c.ywithoutbound
 					//old*0.85+target* 0.15 if diff > 1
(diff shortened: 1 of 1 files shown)


potsmugen:
So far so good. This fixes an issue I'd posted on Discord but not here yet.

Kasasagi77:
OK.

To clarify - I want to continue testing the #1652 issue, but I do think that this PR is an unambigous fix and can be merged even before fully closing #1652.

Eiton:
Looks good. Have you tried using targetScale instead of scale? I think it should be better.

Eiton:
This is not need when #1760 is merged

Kasasagi77:
OK then, I will continue testing when #1760 is merged.

Closing this then.

PR #1788, opened 2024-05-13
feat: InputTime params L/R

Jmorphman:
- Adds two new parameters to the InputTime trigger: absolute Left and Right inputs, which do not change depending on facing.
- InputTime is currently the only way of accessing these absolute directions, since the normal Back/Forward inputs are sufficient in the majority of cases—those who would need Left and Right are going to be using InputTime for what they need anyway.

### DIFF
--- a/src/bytecode.go
+++ b/src/bytecode.go
@@ -527,6 +527,8 @@ const (
 	OC_ex_inputtime_D
 	OC_ex_inputtime_F
 	OC_ex_inputtime_U
+	OC_ex_inputtime_L
+	OC_ex_inputtime_R
 	OC_ex_inputtime_a
 	OC_ex_inputtime_b
 	OC_ex_inputtime_c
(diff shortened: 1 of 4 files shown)


potsmugen:
I have mixed feelings about this one. On one hand, it's adding a harmless feature. On the other, it's adding superfluous code to the inputs for the sake of a minor trigger and for a use case people can already do with existing features.

The original purpose of InputTime (buffer trigger in K4thos's fork) was to eventually unhardcode basic states without relying on commands defined in the cmd files. Myself, I added it because I had a use case where I needed to check for "FF", which is not a required command, so I had to construct it manually. L and R can be checked with B and F plus facing like always. If you specifically need to check how long they were held/released that's also doable with map timers.

EDIT: After checking out that flight code conversation I realized that this *can* address some issues. So in that case it's a net gain and I'm for it.

Jmorphman:
> EDIT: After checking out that flight code conversation I realized that this _can_ address some issues. So in that case it's a net gain and I'm for it.

the flight code conversation, for reference/posterity's sake, from the MFG discord server: https://discord.com/channels/318793724786245633/562828620935069711/1239424332464390194

PR #1789, opened 2024-05-13
feat: Re-enabled FLAC support

Jesuszilla:
go-flac library with Apache License 2.0 added to retrieve total number of samples goflac library with BSD 3-clause License added for decoding Requires libflac 1.4.2 or later (1.4.3 is best) and pkg-config tool to build (on Windows, pkg-config can be installed using `choco install pkgconfiglite`. The archive below contains the .pc files required for pkg-config on Windows (for 32-bit builds, some slight modifications will need to be made to the .pc files to point to Win32).

libflac for Windows files can be found on my server [here](http://www.trinitymugen.net/~Creator/Jesuszilla/xiph_libflac_libs_extract_to_C.7z) (I pulled these files directly from Xiph's releases here on GitHub; libflac is BSD 3-clause).

libflac.dll will need to be placed in the same directory as the IKEMEN-Go EXE on Windows. UNIX (Linux and macOS) should just work as long as `llibflac-dev` is installed.

go.sum has a lot of changes so it may need some review.


K4thos:
closing since we're no longer maintaining our own beep, the new pr should be adopted to work with gopxl/beep

PR #1810, opened 2024-05-25
feat: fightscreenvar(round.called)

wily-coyote:
Returns 1 when the `round<n>` component is currently on screen. Returns 0 otherwise.

I needed to detect when "Round N" was being called and `IsAsserted(Intro)` couldn't reliably detect that; this PR adds a property to FightScreenVar that does exactly that.

### DIFF
--- a/src/bytecode.go
+++ b/src/bytecode.go
@@ -632,6 +632,7 @@ const (
 	OC_ex_fightscreenvar_round_over_wintime
 	OC_ex_fightscreenvar_round_slow_time
 	OC_ex_fightscreenvar_round_start_waittime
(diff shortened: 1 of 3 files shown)


potsmugen:
"round.called" is not a fight screen parameter, so I'm not sure about this one. You can use `IsAsserted(Intro)` but in a loop to see if no character has it.

I also had this idea for a trigger like "RoundStateTime". Combined with FightScreenVar you could check for any phase of the round.

K4thos:
I aslo think RoundStateTime would be more practical. Let's continue the discussion on how to best implement it on discord

PR #1830, opened 2024-06-06
feat: IntroState trigger

wily-coyote:
This PR implements the IntroState trigger from #1172. It allows more precise detection of round states during the intro phase.
Here's a description of the return values for this new trigger:
- `0`: not applicable, or players have gained ctrl after "fight!"
- `1`: same as `RoundState = 0`
- `2`: at least one character is in an intro state, or dialogue is active
- `3`: "round \<n\>" has been called
- `4`: "fight!" has been called

### DIFF
--- a/src/bytecode.go
+++ b/src/bytecode.go
@@ -152,6 +152,7 @@ const (
 	OC_powermax
 	OC_canrecover
 	OC_roundstate
+	OC_introstate
 	OC_ishelper
 	OC_numhelper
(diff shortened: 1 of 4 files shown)


potsmugen:
These should use OC_ex2. OC are for Mugen triggers. OC_ex are for Ikemen (expanded) triggers. OC_ex2 are the same but we just ran out of space in the former.

Issue #335, opened 2021-08-19
Palfx doesn't work with 32-bit png sprites


Rurouni-X:
In MUGEN 1.1, you can use color overlay (palfx) for fonts, in both screenpacks and lifebars. 

Example:
`p1.name.font         = 6,0, 1, 192,163,0`

MUGEN 1.1, P1's name:
![1 1 palfx](https://user-images.githubusercontent.com/68881657/130060381-ce6ea1c7-6b44-4f87-8b2d-4e88228a4b77.png)

Now, with IGO:
![palfxigo](https://user-images.githubusercontent.com/68881657/130060505-ea44dc2c-2f9f-4d29-8822-15ff766fa018.png)

The same situation applies to lifebars sprites, it is not possible to have palfx on sprites used as "front".

Here is the lifebar used in the example:
https://www.mediafire.com/file/q51ot2wkgl3zzwb/MugenClimax.rar/file

PR #1848, opened 2024-06-20
fix: AllPalFx fonts tranparency bug fix

### LINKED ISSUE #1847: Lifebar font transparency is broken when using AllPalFX
### Describe the bug

There is a bug that breaks the transparency of the font in the life bar when using AllPalFX in the current nightly.
This problem occurs in both def and fnt files, so it does not matter what font standard is used.

### To Reproduce

Add the following code to [Statedef -3] in KFM
```
[State -3, AllPalFX]
type = AllPalFX
trigger1 = RoundState = 2
time = 2
add = 50,50,50
mul = 256,256,256
sinadd = 0,0,0,1
invertall = 0
color = 256
```
I made the effect a little brighter for clarity.

### Expected behavior

Maintain font transparency when using AllPalFX as in v0.99 and previous versions

### Screenshots / Video

Without AllPalFX
ikemen096

State using AllPalFX
ikemen095

nightly

### Operating system

Windows

_No response_

Wftk:
PR fixes #1847

### DIFF
--- a/src/font.go
+++ b/src/font.go
@@ -448,7 +448,7 @@ func (f *Fnt) drawChar(
 	if len(f.palettes) != 0 && len(f.coldepth) > int(bank) &&
 		f.images[bt][c].img[0].coldepth != 32 &&
 		f.coldepth[bank] != f.images[bt][c].img[0].coldepth {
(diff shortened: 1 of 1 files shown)


Wftk:
@K4thos seems like it also fixes #335

Issue #170, opened 2020-09-06
Stage issue: 4:3 stage on wide resolution


K4thos:
In mugen if we use 16:9 resolution (e.g. 1280x720) and stage with 4:3 localcoord (e.g. 320x240) then by default stage is rendered in 4:3 aspect ratio:
![To7NKTd](https://user-images.githubusercontent.com/8928195/92315528-178fb800-efe7-11ea-89c1-c93b2d4f3c12.png)

Ikemen Go renders such stage using 16:9 aspect ratio, zooming it in:
![1AWuQQY](https://user-images.githubusercontent.com/8928195/92315533-32fac300-efe7-11ea-8074-79a5897638b6.png)

There is no such problem when the stage uses the same aspect ratio as game resolution.

Additional notes: Current Ikemen GO behaviour actually also exists in mugen after disabling stagefit option in mugen.cfg, but since the default mugen 1.1 option is "stage shrunk to fit into screen", and the fact that drawning to width of screen causes visual bugs on many 4:3 stages, we should consider the way it works now as compatibility problem with mugen resources (once solved, having an option for current behaviour as either config.json setting or stage parameter would be the best solution).

Shadic13:
Is there any updates on this or are you guys focusing on getting the other issues fixed and saving this feature for the 1.0 version of Ikemen.

Serasul1984:
this did not change, you cant play in 16:9 without the engine zooming in

K4thos:
That's why the ticket is open. None of the active devs have shown interest in this Mugen 1.1 feature.

arroyoquiel:
> this did not change, you cant play in 16:9 without the engine zooming in

I'm interested in giving a shot to this fix, I love 16:9 screenpacks, but I want stages to be in 4:3

K4thos:
sure, go ahead, we're open for contributions

PR #1868, opened 2024-07-02
feat: Resizable window, scale game to fit window

SuperFromND:
https://github.com/ikemen-engine/Ikemen-GO/assets/22881403/40b5c982-a8e3-41ab-9de2-8dd14d898375

This change re-enables the ability to resize the game window and allows Ikemen GO to scale the game's viewport to fit the window by using a new function, GetScaledViewportSize().

`GetScaledViewportSize()` is a basic algorithim that just scales the GameWidth/GameHeight to the size of the window while maintaing aspect ratio, and then applies this to the viewport with a `gl.Viewport` command. Sadly the result doesn't apply any filter when scaling, so you get nearest-neighbor scaled pixels, but it's at least better than no resizing the window at all IMO.

Window size is not saved, and GameWidth/GameHeight still determine the starting size of the game's resolution and window size.

I hope to give this more flexibility in the future, such as allowing the end-user to toggle the aspect ratio correction and maybe integer scaling as well.

### DIFF
--- a/src/render_gl.go
+++ b/src/render_gl.go
@@ -421,6 +421,8 @@ func (r *Renderer) EndFrame() {
 		gl.BlitFramebuffer(0, 0, sys.scrrect[2], sys.scrrect[3], 0, 0, sys.scrrect[2], sys.scrrect[3], gl.COLOR_BUFFER_BIT, gl.LINEAR)
 	}
 
+	x, y, resizedWidth, resizedHeight := sys.window.GetScaledViewportSize()
+	gl.Viewport(x, y, resizedWidth, resizedHeight)
 	gl.BindFramebuffer(gl.FRAMEBUFFER, 0)
 
 	postShader := r.postShaderSelect[sys.postProcessingShader]
--- a/src/system_glfw.go
+++ b/src/system_glfw.go
@@ -35,7 +35,7 @@ func (s *System) newWindow(w, h int) (*Window, error) {
 	_, forceWindowed := sys.cmdFlags["-windowed"]
 	fullscreen := s.fullscreen && !forceWindowed
 
-	glfw.WindowHint(glfw.Resizable, glfw.False)
+	glfw.WindowHint(glfw.Resizable, glfw.True)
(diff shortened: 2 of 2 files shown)


K4thos:
Can this method, or some variant of it, be used to implement these missing Mugen 1.1 settings when integrated further? (They didn't exist in Mugen 1.0 and are possible thanks to Mugen 1.1 OpenGL RenderMode.)

From mugen.cfg
 ;The video resolution defaults to the same as the game resolution.
 ;In windowed mode, this specifies the window size.
 ;You can force an alternate resolution by uncommenting the lines
 ;below.
;Width  = 1280
;Height = 720

 ;Set to 1 to make the window resizable when in windowed mode.
 ;Only supported for RenderMode=OpenGL and OpenGLScreen.
Resizable = 1

 ;Set to 0 to stretch the video to fit the whole window.
 ;Set to 1 to keep a fixed aspect ratio.
KeepAspect = 1

 ;Stage fit mode.
 ;0 - stage drawn to width of screen (may crop stages with tall aspect)
 ;1 - stage shrunk to fit into screen
 ;Ignored if RenderMode=System (forced to 0)
StageFit = 1

 ;System fit mode.
 ;0 - system drawn to width of screen (may crop motifs with tall aspect)
 ;1 - system shrunk to fit into screen
 ;Ignored if RenderMode=System (forced to 0)
SystemFit = 1

Probably the most anticipated missing Mugen 1.1 feature is StageFit = 1 support, since that's the default Mugen 1.1 behavior. (Mugen 1.0 works like Ikemen GO in this regard.)
https://github.com/ikemen-engine/Ikemen-GO/issues/170

SuperFromND:
`KeepAspect` is pretty trivial to support as that just entails skipping most of the `GetScaledViewportSize` function. `StageFit` and `SystemFit` might not be possible here since we're not updating the actual _game_ resolution.

I'm not really sure if `Resizable` is even desirable to have as a toggle as having it be one just seems like asking for false bug reports. LOL

PR #1872, opened 2024-07-06
feat: Added ability to customize behavior of master volume during menu pause.

Jesuszilla:
Addresses feature request #1234

* Added PauseMasterVolume to config (defaults to 0: current behavior)
* Music pauses if master volume during pause low enough (< 5); scaled by configged value otherwise.
* All sounds now muted by the pause master volume amount.


K4thos:
please also update the defaukt config file: https://github.com/ikemen-engine/Ikemen-GO/blob/develop/src/resources/defaultConfig.json

PR #1883, opened 2024-07-11
Update README.md

shinriyo:
typo the name of `docker_build.sh`


K4thos:
from what I see this change was already made

PR #1907, opened 2024-07-25
feat: WindowScaleMode (nearest/bilinear filter toggle)

SuperFromND:
Requested by a few people and trivially easy to implement, so why not?
Defaults to `true`, bilinear filtering, while setting it to `false` uses nearest-neighbor.
image
image

### DIFF
--- a/src/main.go
+++ b/src/main.go
@@ -221,6 +221,7 @@ type configSettings struct {
 	InputSOCDResolution        int32
 	IP                         map[string]string
 	KeepAspect                 bool
+	WindowScaleMode            bool
 	LifeMul                    float32
 	ListenPort                 string
 	LoseSimul                  bool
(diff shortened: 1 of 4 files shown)


Kasasagi77:
Hello @SuperFromND - I wanted to throw in a scaling feature idea, but please treat this comment as an idle discussion for Your consideration rather than a full feature request.

Some software dealing with low-res/pixel-art scaling also adds an option sometimes called "sharp bilinear". The concept is very simple - you first upscale the original image with the "nearest" filter by a configurable integer ratio, e.g. x2, x3, x4, x5 and so on, and then scale the result with "bilinear" filter to the screen size (which can also be a down-scaling when a large multiplier is used).

This metod is very simple, but the results are great in my opinion.

This method can also often eliminate the motion-shimmer artifacts characteristic to the "nearest" scaling without sacrificing too much sharpness.

This method is also a strict generalization of the normal bilinear filtering, so it could be controlled with only one parameter in the configuration, e.g. `BilinearPrescale`, with `0` meaning only "nearest", `1` meaning only "bilinear", `>2` meaning the "nearest" pre-scaling factor of this method.

SuperFromND:
I'm actually familiar with sharp-bilinear, I'm just not entirely sure how to implement it. IIRC RetroArch has a working implementation I could probably look too, but for right this second its not a huge priority for me.

PR #1913, opened 2024-07-27
fix: Various window resizing-related fixes

### LINKED ISSUE #1901: When opening ikemen go nightly even with msaa disabled i get a black screen
I get a black screen when I open ikemen go nightly with msaa disabled

SuperFromND:
Fixes a regression that caused Ikemen GO to crash when attempting to examine it via Renderdoc, and ~~suspected to also be the cause of~~ permanent black-screens on Intel Integrated graphics due to undefined behavior (binding a texture as if it were a framebuffer object) that Nvidia and AMD gloss over with their drivers.

I suggest merging this before merging my other PR (#1907) as this fixes a fairly severe regression (~~even if this isn't the fix for Intel~~, it at least fixes Renderdoc which is essential for graphics debugging) while the other adds a (albeit very minimal-code) feature.

UPDATE: I just got confirmation this fixes Intel integrated graphics having a black screen! Fixes #1901.

### DIFF
--- a/src/render_gl.go
+++ b/src/render_gl.go
@@ -446,14 +446,20 @@ func (r *Renderer) EndFrame() {
 	gl.DrawArrays(gl.TRIANGLE_STRIP, 0, 4)
 	gl.DisableVertexAttribArray(uint32(loc))
 
-	// resize viewport and scale finished frame to window
-	gl.Viewport(x, y, resizedWidth, resizedHeight)
+	// rebind to prepare frame for blitting to window
 	if sys.multisampleAntialiasing {
 		gl.BindFramebuffer(gl.READ_FRAMEBUFFER, r.fbo_f)
 	} else {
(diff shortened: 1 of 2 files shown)


SuperFromND:
Appended a second commit to this that also fixes a related problem where resizing the window causes garbage data to appear due to us not clearing the entire window, just the frame area. Hope that's okay!

K4thos:
Please resolve the conflict, I'm merging from phone

SuperFromND:
Should be good to go now 👍

shin258:
> Should be good to go now 👍

Hi I confirm, it works great in the screen pack and fight with windowScalemode and Keep-aspect option activated also is corrected some frames drop too in my case (due of fbo bug before this last build i think). many thanks for your work. 
Menu: (no systemfit and stagefit option)
menu
In fight:
windowscale

PR #1924, opened 2024-08-03
fix: Re-use palette textures, evict SFF cache entries

worleydl:
Reusing palette textures prevents flooding of garbage collection.  Evicting SFFCache entries prevents clawing of RAM over time.

### DIFF
--- a/src/anim.go
+++ b/src/anim.go
@@ -126,6 +126,7 @@ type Animation struct {
 	sff                *Sff
 	palettedata        *PaletteList
 	spr                *Sprite
+	tex		   *Texture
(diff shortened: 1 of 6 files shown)


worleydl:
Converting to draft until stability issues are resolved.

worleydl:
Closing because it appears these were only issues when running in c-shared mode which seems to cause Finalizers to not be triggered.  It may be good practice in the future to re-use palette textures but need to figure out an option that runs on the main platforms (I was unable to replicate crashes seen by others with this code if someone wants to jot down their environment and the screenpacks used I can give it another whirl)

PR #1931, opened 2024-08-06
fix: Swap & bind framebuffer when taking screenshots, skip viewport resize on window/game res being the same, add MSAA samples setting

### LINKED ISSUE #1922: Taking screenshots results in corrupted/not-ideal images with a resized window when MSAA is enabled
### Describe the bug

Taking a screenshot with a window at a resolution other than the internal game resolution results in the screenshot either A. being only the game's internal resolution and the rest filled with solid black, or even worse B. corrupted in addition to A.

Examples of such screenshots can be found below.

### To Reproduce

1. Download the latest nightly build and run it.
2. Screenshot the game with F12.
3. Observe resulting screenshots

### Expected behavior

The game's screenshots should either A. be identical to what the window shows or B. be in native resolution. I'm going to assume the former is more desirable here.

### Screenshots / Video

ikemen003
ikemen045
ikemen046

Latest Nightly (7/31/2024)

### Operating system

Windows

screenshot glitch corruption nightly window resize

SuperFromND:
Fixes #1922.

For a more detailed explanation of this bug, see my replies to that issue. _Hopefully_ this is the last commit I have to make regarding screen resize regressions for a while!

### DIFF
--- a/data/system.base.def
+++ b/data/system.base.def
@@ -1638,8 +1638,11 @@
 	menu.itemname.menuvideo.resolution.customres = "Custom"
 	menu.itemname.menuvideo.resolution.back = "Back"
 	menu.itemname.menuvideo.fullscreen = "Fullscreen"
+	menu.itemname.menuvideo.keepaspect = "Keep Aspect Ratio"
+	menu.itemname.menuvideo.windowscalemode = "Bilinear Filtering"
 	menu.itemname.menuvideo.vretrace = "VSync"
 	menu.itemname.menuvideo.msaa = "MSAA"
+	menu.itemname.menuvideo.msaasamples = "MSAA Samples"
 	menu.itemname.menuvideo.shaders = "Shaders" ;reserved submenu
 	; This list is populated with shaders existing in 'external/shaders' directory
 	menu.itemname.menuvideo.shaders.empty = ""
--- a/data/system.base.def
+++ b/data/system.base.def
@@ -1642,7 +1642,6 @@
 	menu.itemname.menuvideo.windowscalemode = "Bilinear Filtering"
 	menu.itemname.menuvideo.vretrace = "VSync"
 	menu.itemname.menuvideo.msaa = "MSAA"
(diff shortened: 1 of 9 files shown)


SuperFromND:
Context for the other two commits:

[7e04f0e](https://github.com/ikemen-engine/Ikemen-GO/pull/1931/commits/7e04f0e42e2394435eb4a2ca1412cfd74026b268) adds a second condition to skip scaling the viewport if the game's window and game resolution are exactly the same. This fixes an issue that caused blurring on certain resolutions (384x224 for example) due to floating point imprecisions.

[bd7661e](https://github.com/ikemen-engine/Ikemen-GO/pull/1931/commits/bd7661e21174919271a999628864df64f9c93a66) adds a new config option, `MSAASamples`, that allows the user to configure how many samples MSAA should use. It defaults to 2, which is notably much lower than our previously-hardcoded 16, but it appears this value allows MSAA to work on certain platforms that otherwise produced black screens when attempting to use it (MacOS machines, for example; thanks @Jesuszilla for testing that for me!).
ikemen MSAA comparison

SuperFromND:
Alright, last commit on this PR since it's getting pretty cluttered. This one just adds all of the options I've added recently into the video settings and properly adds them to the Lua config so that default values will be set.
ikemen048

K4thos:
I think it's better to merge MSAA and MSAA samples config options into one, like in most games. So that would be MSAA: Disabled/x2/x4/x8/x16. What do you think, Super? Looks like there are also some merge conflicts after merging localization pr

SuperFromND:
> I think it's better to merge MSAA and MSAA samples config options into one, like in most games. So that would be MSAA: Disabled/x2/x4/x8/x16. What do you think, Super? Looks like there are also some merge conflicts after merging localization pr

I agree, but it might cause some issues since my current default is to set MSAA samples to 2, but I'll see what I can do!

SuperFromND:
Should be good to go now! Worth noting that this change breaks older configs since MSAA was previously a boolean rather than an int32, so whenever next stable comes around, take note of that.

PR #1939, opened 2024-08-08
fix(motif): language merging behavior

wily-coyote:
- Language-related sections are also merged with the screenpack's defaults after the rest of the motif is parsed. Previous behavior would only merge with motif.lua defaults.

- Fixed a Lua error that would occur when a language-related section doesn't have a `menu.pos` parameter.


Orange-Dolphin:
I want to make a few brief mentions regarding this. When I was creating the functionality, I was focused on matching Mugen behavior. Mugen did complete sections as opposed to only taking what has changed. I think doing it the way this PR has it set up is much better. You can change the name of a menu option without having to replace the entire [Title Info] block for example. That isn't how it was done in mugen, but it's a lot less bloated, and should still fit any screenpacks or features that had these features before.

So although it doesn't match Mugen behavior 1:1, I support this way of going about it.

PR #1950, opened 2024-08-16
fix: 3D stage rendering and postprocessing shaders on macOS, new trigger redirects

Jesuszilla:
* Fix 3D stages not rendering on macOS (Intel and ARM).
* Fix postprocessing shaders not working on macOS (Intel and ARM).
* Fix redirects on explodVar, projVar, and clsnVar not working properly.

This will require updates to the glfont repo in order to work properly. macOS (Intel and ARM both) can only support a minimum OpenGL version of 3.2 and must make use of a VAO for proper 3D rendering hence the need for these changes. Shaders were rewritten with compatibility macros.

Huge thanks goes to assemblaj/fantasma for the initial 3D stage rendering fix with the VAO!

### DIFF
--- a/external/shaders/HQ2x.frag
+++ b/external/shaders/HQ2x.frag
@@ -1,3 +1,15 @@
+#if __VERSION__ >= 130
+#define COMPAT_VARYING in
+#define COMPAT_TEXTURE texture
+out vec4 FragColor;
+in vec4 TexCoord[7];
+#else
+#define COMPAT_VARYING varying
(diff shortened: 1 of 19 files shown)


K4thos:
merged in glfont repo, please update go.mod and go.sum

Jesuszilla:
> merged in glfont repo, please update go.mod and go.sum

Done! Also changed clsnVar left to "back", and "right" to "front" so it's more intuitive.

PR #1986, opened 2024-09-04
feat: Universal Trials Mode

two4teezee:
A massive PR that basically takes this: https://github.com/two4teezee/Ikemen-GO-Trials-Mode and merges it into the core engine. Includes all necessary changes to the default config file, system.base.def, etc. Unsurprisingly, given how extensive the mode is, it touches many aspects of the existing Lua files. The only real gains from delivering the mode with the engine would be...

1) Greater visibility and accessibility to the mode
2) A small (extremely small) efficiency gained by not having to open the char .def files in Lua after the engine has already parsed them.
3) No duplicate code between training.zss and trials.zss


two4teezee:
If this gets merged in, should also merge this: https://github.com/ikemen-engine/Ikemen_GO-Elecbyte-Screenpack/pull/20

external/script/trials.lua
```diff
@@ -0,0 +1,906 @@
...
+		end
+	end
+end
+
+-- Find trials files and parse them; append t_selChars table
+function trials.f_parseTrials(row)
+	i = 0 --Trial number
+	j = 0 --TrialStep number
+	trial = {}
+	local trialsFile = main.f_fileRead(main.t_selChars[row].trialspath, "r")
```
bluish-elvis: ~~If Character has not trials.def, IKEMEN will crash due to not found trials.def.
It looks like necessary to check it by main.f_fileExists~~

two4teezee: You aren't wrong but this function is only invoked by start.lua if trialspath ~= "" - in other words, it's only resolved when a trials file was found when reading the char def file.

bluish-elvis: also I tested this build and your sample of KFM trials.def but panicError caused on this f_fileRead even there IS trials.def...
```suggestion
	local trialsFile = main.f_fileRead(main.t_selChars[row].dir .. main.t_selChars[row].trialspath, "r")
```

bluish-elvis: also I tested this build and your sample of KFM trials.def but panicError thrown by this f_fileRead even there IS trials.def...
```suggestion
	local trialsFile = main.f_fileRead(main.t_selChars[row].dir .. main.t_selChars[row].trialspath, "r")
```

two4teezee: Thanks, it resolved the path in MacOS, but you're right, what I had wasn't working in x64. I accepted the change, try it again @bluish-elvis

two4teezee: This change was breaking for MacOS but I included a change that leveraged your first suggestion as well.

bluish-elvis: Confirmed it's work. thanks!

PR #2000, opened 2024-09-14
feat: Collisions Rotation feat and Text sctr/Debug text blend mode inverse bug fix

### LINKED ISSUE #1091: True type fonts are bugged in layerno 1
### Describe the bug

For some reason if you draw true type fonts (tested with debug font) in layer 1 in a stage that uses a subtractive layer for a dark effect, the fonts will be black.

### To Reproduce

1. Download the stage from #1088 
2. Put this in KFM's negative states:
```
[State -2, Test]
type = text
trigger1 = !(gametime % 60)
removetime = 30
layerno = 1
text = "TEST"
pos = 160, 120
ignorehitpause = 1
```
3. The text should be white, but in that stage it is black
4. If you change the layerno parameter the issue disappears

### Expected behavior

_No response_

### Screenshots / Video

ikemen013

0.98.2 and 0.99 RC1

### Operating system

Windows

_No response_

### LINKED ISSUE #1219: EX Yu: Black Debug text
### Describe the bug

For some reason, EX Yu seems to be making the debug text turn black when he's on screen, sometimes it shifts back to white, but it's usually only for a bit before shifting back to black. Checked on Mugen and it doesn't seem to change to black no matter what.

### To Reproduce

1: Download EX Yu here (https://drive.google.com/file/d/1-hQn-swYhJdnOt0VWL0oBw6PKNJ1qn-R/view)
2: Start a match with EX Yu and press CTRL+D to show debug text
3: It should turn black as the shadowy effects start appearing during the char intro.

### Expected behavior

The Debug text shouldn't be turning black i'm guessing.

### Screenshots / Video

EXYu 1
Look at top left and bottom left.

Nightly Build 5/6/23

### Operating system

Windows

_No response_

Wftk:
Feat:

- Now Collision boxes can be rotated.In that case: **rotateclsn**, **projrotateclsn** parameters been added to Angledraw and Projectile sctrls. For enabling value need to be set to 1.

Fix:

- Text sctr/Debug text blend mode inverse bug fix #1091

### DIFF
--- a/src/anim.go
+++ b/src/anim.go
@@ -944,6 +944,7 @@ func (dl DrawList) draw(x, y, scl float32) {
 		}
 		sys.brightness = ob
 	}
+	BlendReset()
 }
 
 type ShadowSprite struct {
@@ -1083,7 +1084,6 @@ func (sl ShadowList) drawReflection(x, y, scl float32) {
 					(s.pos[1]-s.reflectOffset[1]-offsetY), scl, scl, s.scl[0], s.scl[0], -s.scl[1]*sys.stage.reflection.yscale, 0,
 				s.rot, float32(sys.gameWidth)/2, s.fx, s.oldVer, s.facing, true, s.posLocalscl, s.projection, s.fLength)
 		}
-
 	}
 }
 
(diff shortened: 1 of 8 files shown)


K4thos:
Deployment failed on macos
Error: src/render.go:288:6: gfx.BlendReset undefined (type *Renderer has no field or method BlendReset)
https://github.com/ikemen-engine/Ikemen-GO/actions/runs/10861509608/job/30143494576

PR #2032, opened 2024-09-28
feat: gethitvar(guardflag)

wily-coyote:
This PR adds a `guardflag` parameter to the `gethitvar` trigger.

Using this parameter with the trigger returns a bitflag; to check for a specific type, use the `BitN` constants:

- `Const(Bit1)`: High
- `Const(Bit2)`: Low
- `Const(Bit3)`: Air

Mid is equivalent to `Const(Bit1)|Const(Bit2)`.

### DIFF
--- a/src/bytecode.go
+++ b/src/bytecode.go
@@ -509,6 +509,7 @@ const (
 	OC_ex_gethitvar_down_recover
 	OC_ex_gethitvar_down_recovertime
 	OC_ex_gethitvar_xaccel
+	OC_ex_gethitvar_guardflag
 	OC_ex_ailevelf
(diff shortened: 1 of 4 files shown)


potsmugen:
This isn't how MoveHitVar works. It should set the variables on contact like GetHitVar. For the implementation you're doing I've been thinking about a HitDefVar trigger.

wily-coyote:
I'll get rid of `guardflag` on `MoveHitVar` for now since it isn't really what I need for the project I'm working on right now.
Perhaps if a HitDefVar trigger does get made, it'll have `guardflag` as one the parameters.

PR #2057, opened 2024-10-06
fix: explodVar interpolation with scale and angle

Jesuszilla:
Fixes interpolation not being taken into account for scale and angle in explodVar.


Jesuszilla:
Forgot to do script.go, will be back

PR #2077, opened 2024-10-14
fix: Truncate RenderParams X/Y before rendering

SuperFromND:
This PR eliminates a rather minor but annoying blur that occurs when running Ikemen with MSAA enabled and using low-res characters, then proceeding to move around said characters.

image

The reason this happens is due to Ikemen using floating-point coordinates for X, Y, width, and height when rendering most things to the screen. This won't really be an issue to most users, but for certain edge cases it can rear its ugly head, especially if a creator's intent is to have nice sharp pixels that rely on perfectly even coordinates.

HD screenpacks suffer from a related issue when attempting to use pixel-perfect assets at certain localcoord scales like 1280x720; this fix is part of my attempts to eventually rectify this per request of the author of one such screenpack.

The strange math here is simply because I didn't want to add an extra "math32" library just to truncate these 32-bit float values.

### DIFF
--- a/src/render.go
+++ b/src/render.go
@@ -276,6 +276,10 @@ func rmInitSub(rp *RenderParams) {
 		rp.rot.angle *= -1
 		rp.rot.xangle *= -1
 	}
+
+	rp.x = float32(math.Trunc(float64(rp.x)))
+	rp.y = float32(math.Trunc(float64(rp.y)))
+
 	if rp.tile.x == 0 {
 		rp.tile.sx = 0
 	} else if rp.tile.sx > 0 {


SuperFromND:
Converted to draft as there's been some differing opinions on this. Might make this a toggleable option since it's not always desired for certain things like slowly-scrolling backgrounds.

SuperFromND:
Closing this since I think it's a much better idea to try and find where character rendering is called from and apply the rounding fix to just that, rather than the entire rendering function.

Also, this was pushed onto a master branch and I need to push it to a different branch anyways LOL

PR #2083, opened 2024-10-19
fix: issue where unset hitdef ID's can trigger TargetStates with specified ID parameter

Jesuszilla:
Fixes an issue where unset HitDef ID's can trigger TargetStates that have specified ID parameter.


Jesuszilla:
Changed to draft because the issue now seems to point towards cond exploit not working as intended actually?

Jesuszilla:
fixed by #2085

PR #2090, opened 2024-10-26
fix: palette select issues

rakieldev:
Fixes: 
- Palettes will no longer jump from 9 to 4 or 7 to 8 when using Palette Keymaps; all palettes now follow in order
- Removed the limit of cycling through only 6 palettes without holding Start; palettes will now cycle through all available
- Fixed an issue where, when exceeded, palettes would randomize unexpectedly due to the palette keymap

### DIFF
--- a/external/script/start.lua
+++ b/external/script/start.lua
@@ -605,15 +605,8 @@ function start.f_setMusic(num, data)
 	end
 end
 
---remaps palette based on button press and character's keymap settings
-function start.f_keyPalMap(ref, num)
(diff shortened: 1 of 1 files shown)


Kiwi1444:
I'm disagree, that now palettes cycled through pal 11 and pal 12, as my character (and many other characters) has special modes activated by choosing that palettes. And it is not assumed that these palettes can be chosen without holding start. This new behavior can only take place if all other palettes are already chosen. It would be logical that the palettes are cycling through what is specified in `pal defaults`.

PR #2095, opened 2024-11-02
feat: Simple Environment map, punctual light and shadow

Eiton:
Add support for environment map and punctual light in 3d stage.
New stage def parameters:
```
environment: .hdr file for environment map
environmentintensity: change the environment intensity. Default is 1.0
```
New glTF custom parameters:
```
shadowMapBias: default value is 0.02
shadowMapNear: deafault value is 0.1 for point / spot light, -20 from directional light
shadowMapFar: default value is 50
shadowMapBottom: directional light only, default value is -20
shadowMapTop: directional light only, default value is 20
shadowMapLeft: directional light only, default value is -20
shadowMapRight:  directional light only, default value is 20
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -26,6 +26,9 @@ require (
 	github.com/hajimehoshi/go-mp3 v0.3.4 // indirect
 	github.com/jfreymuth/oggvorbis v1.0.5 // indirect
 	github.com/jfreymuth/vorbis v1.0.2 // indirect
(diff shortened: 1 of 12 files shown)


K4thos:
not sure why the linux build fails
https://github.com/ikemen-engine/Ikemen-GO/actions/runs/11658706774/job/32458148282

PR #2199, opened 2024-12-22
feat: mugen style config.ini, GameOption trigger

K4thos:
- Added a Mugen-style configuration file with comments.
- AI ramping settings have been moved from `select.def` to `config.ini`, similar to Mugen.
- Deprecated the use of `config.json`.
- The `GameOption` trigger now returns all available options.
- Enhanced functionality in `iniutils.go`  supporting upcoming updates for motif and storyboard parsing.
- Refactor: Removed 33 functions from script.go and 60 variables from the System struct. All game option data is now grouped under a single struct, enabling read/write functionality accessible from both the engine and external Lua scripts.


two4teezee:
@K4thos can you explain how `gamespeed` maps back to framerate? I reviewed the changes but the relationship between the two wasn't immediately obvious to me and this is a massive PR.

K4thos:
game still runs at 60 fps even if you adjust gamespeed. What controls actual FPS is Framerate under [Config] but I'm not sure if we should even expose this configuration considering it also affects speed at which menus are endered, so only 60 is fully supported.

PR #2255, opened 2025-01-16
fix: scaledelta is affected by zoom

### LINKED ISSUE #2228: Incorrect (not matching Mugen 1.1) background scaling with `scaledelta` and zoom<>`1.0`
### Describe the bug

The scale/size of the backgrounds using `scaledelta` in a stage also using zoom other than 1.0 (even a single constant zoom value) does not match Mugen 1.1.

### To Reproduce

Grab a test stage from [here](https://drive.proton.me/urls/Y7P997G2HR#3ubNsbgBjHnT).

#### Scenario A

1. Use the `NewOffice-zoom-mugen.def` stage from the above archive.

2. Note that when a character jumps in _Ikemen_,
    a. there is a gap between the floor and the window frame,
    b. there is a gap between the desk and the floor shadow.

3. The above gap is larger for a smaller zoom value (i.e. a bigger zoom-out).

4. Note that when a character jumps in _Mugen 1.1_:
    a. the floor and the window frame are perfectly aligned, regardles of zoom,
    b. the desk and the floor shadow is perfectly aligned, regardles of zoom.

#### Scenario B

1. Use the `NewOffice-wide-mugen.def` stage from the above archive.

2. Note that this stage has a contant zoom of exactly `0.75`.

3. Note that, in Ikemen, there are gaps between backgrounds like in **Scenario A**.

4. Note the `scaledelta` value for the floor background:
```
[BG Fighting Ground]
scaledelta = 0, -0.00435
```

5. Note that setting the `scaledelta` to the value of `0, -0.0058` instead compensates for the gaps problem in Ikemen and that (probably not coincidently):
```
-0.0058 * 0.75 = -0.00435
```

6.  Note that there is no way to compensate for this problem in a stage where zoom is an actual range (as opposed to a single value) as the alignment will only be correct for at most a single value of zoom, i.e. there is currently not even a proper work-around for this bug in Ikemen.

### Expected behavior

The behaviour should match _Mugen 1.1_.

In the context of the above testing scenarios, the floor and the window frame should be perfectly aligned when a character jumps, regardles of the current zoom, i.e.:

- there should be **no gaps** between the floor and the window frame,

- there should also be **no overlap** between the floor and the window frame<br/>(i.e. the floor should not slide under the window frame).

#### Ikemen / Problem
ikemen

#### Mugen 1.1 / Correct
mugen

2024-11

### Operating system

Linux

The solution is most probably to apply a current zoom correction to the `scaledelta` value of backgrounds.

NeatUnsou:
- scaledelta is affected by zoom
Fixes #2228
- Fixed a typo

### DIFF
--- a/src/stage.go
+++ b/src/stage.go
@@ -403,7 +403,7 @@ func (bg backGround) draw(pos [2]float32, drawscl, bgscl, stglscl float32,
 	y := bg.start[1] - yScrollPos + bg.bga.offset[1]
 
 	// Calculate Y scaling based on vertical scroll position and delta
-	ys2 := bg.scaledelta[1] * pos[1] * bg.delta[1] * bgscl
+	ys2 := bg.scaledelta[1] * pos[1] * bg.delta[1] * bgscl / drawscl
 	ys := ((100-(pos[1])*bg.yscaledelta)*bgscl/bg.yscalestart)*bg.scalestart[1] + ys2
 	xs := bg.scaledelta[0] * pos[0] * bg.delta[0] * bgscl
 	x *= bgscl
--- a/src/bytecode.go
+++ b/src/bytecode.go
@@ -1791,7 +1791,7 @@ func (be BytecodeExp) run(c *Char) BytecodeValue {
 		case OC_pos_x:
 			sys.bcStack.PushF((c.pos[0]*(c.localscl/oc.localscl) - sys.cam.Pos[0]/oc.localscl))
 		case OC_pos_y:
-			sys.bcStack.PushF((c.pos[1] + c.groundLevel - c.platformPosY) * (c.localscl / oc.localscl))
+			sys.bcStack.PushF((c.pos[1] - c.groundLevel - c.platformPosY) * (c.localscl / oc.localscl))
 		case OC_power:
 			sys.bcStack.PushI(c.getPower())
 		case OC_powermax:


src/stage.go
```diff
@@ -403,7 +403,7 @@ func (bg backGround) draw(pos [2]float32, drawscl, bgscl, stglscl float32,
 	y := bg.start[1] - yScrollPos + bg.bga.offset[1]
 
 	// Calculate Y scaling based on vertical scroll position and delta
-	ys2 := bg.scaledelta[1] * pos[1] * bg.delta[1] * bgscl
+	ys2 := bg.scaledelta[1] * pos[1] * bg.delta[1] * bgscl / drawscl
 	ys := ((100-(pos[1])*bg.yscaledelta)*bgscl/bg.yscalestart)*bg.scalestart[1] + ys2
 	xs := bg.scaledelta[0] * pos[0] * bg.delta[0] * bgscl
```
Kasasagi77, line 408: @NeatUnsou 

I can't help but notice that this code implies that the `y` dimension of `scaledelta` is affected by zoom, but the `x` dimention is not. Is that actually the case? Does Mugen 1.1 behave this way? Or do I not understand this code correctly?

NeatUnsou: After making it so that it also affects the x dimension, I confirmed a stage where a bug occurs.
Yes, I agree that it doesn't seem correct intuitively, but after testing with around 30 stages that use zoom and scaledelta, I couldn't find any issues, so that's why it ended up like this.
If you find any stages where bugs occur related to scaledelta, please create an issue.

Kasasagi77: Great! Thank you very much 🙏.

PR #2288, opened 2025-01-29
fix: default DepthToScreen, ProjCancelTime, invisible transparency, other

### LINKED ISSUE #2279: ProjCancelTime nonworking function
### Describe the bug

[kfmp.zip](https://github.com/user-attachments/files/18550588/kfmp.zip)
This KFM demonstrates projectiles and at least with a simple purpose, ProjCancelTime. In his -3 statedef, I want them to play an alt noise when it detects and collides at another projectile.
This was more like a missing feature request, but that template was changed and such matters go to the Github discussions now.

### To Reproduce

1. Load kfmp for P1 & P2 both.
2. Crouch for both players, hit X+Y for both players. Introductory wood board is output as projectile with clsn1 and clsn2.
3. Have two thrown kfm wood boards collide, hear the alt noise going 'HAH'.

### Expected behavior

"Projectile collided with another projectile" unique 'canceled' noise, or as I demonstrate through what's available on KFM himself, his HAH sound S5,0 to come out.  This currently works only in MUGEN 1.0+.

> This trigger takes an required nonnegative ID number as an argument. If the player's last projectile to make any kind of contact was cancelled by an opponent's projectile and had the specified ID number, then ProjCancelTime returns the number of ticks since that contact occurred. If the specified ID number is 0, then the projectile ID is not checked. If no projectile meets all the above conditions, then ProjCancelTime returns -1.
> 
> Format:
> ProjCancelTime(exprn)
> 
> Arguments:
> exprn
> Expression evaluating to a nonnegative ID number (int).
> 
> Return type:
> int
> 
> Error conditions:
> Returns bottom if exprn evaluates to bottom. If a negative ID is specified, then the ID defaults to zero.

### Screenshots / Video

https://github.com/user-attachments/assets/4f47e3fb-01c2-4949-b98a-0737a2eac528
Ikemen ProjCancelTime (nought)

https://github.com/user-attachments/assets/d61483a5-8691-4cf8-a542-9ff8733c6adc
MUGEN 1.0 ProjCancelTime (forgive the resolution capture disparity)

Seemingly all IKEs

### Operating system

Windows

[Projectile Sctrl info](https://mugenguild.com/forum/topics/projectile-sctrl-169755.0.html)
[ProjCancelTime standard info](https://mugenguild.com/forum/PHPSESSID.8sicthbjin082eggq4of6l12us/topics/projcanceltime-triggers-169922.0.html)
[Other ProjCancelTime info](https://levelalfaomega.gitbook.io/unity-mugen/unity-mugen/trigger-list/projcanceltime)

### LINKED ISSUE #2281: Transparency SCTRL inconsistency between Mugen and Ikemen
### Describe the bug

The transparency SCTRL (with addalpha) does not behave the same way in Ikemen as it does in Mugen. In Ikemen, if the first alpha value reaches 0 or lower, the character with this SCTRL will disappear from the screen, but in Mugen, the character will remain on screen but turn all black.

The following block of code is what causes the situation shown below:

[State 0, Trans]
type = Trans
trigger1 = 1
trans = addalpha
alpha = 256-time*10,0

### To Reproduce

Use Ichida/Vinnie's Blake Belladonna and perform her Astral Heat (down twice then any two kick buttons, or D, D + a/b/c on an opponent with at most 1/4 of their max health remaining).

The last part of her Astral Heat normally (i.e. in Mugen) has her and the opponent turn all black, and then the opponent slowly fades away. But here, once the background turns red, Blake disappears when she's not supposed to. The opponent also vanishes instead of fading away due to having this SCTRL in the custom state they are put into.

### Expected behavior

In Mugen, it looks like this: 

https://github.com/user-attachments/assets/d46d0775-eec5-4ec4-9790-283ed3903843

### Screenshots / Video

But in Ikemen, it looks like this:

https://github.com/user-attachments/assets/cd8bc351-79de-4f6f-948e-28fa91ba909c

nightly

### Operating system

Windows

_No response_

potsmugen:
Fix:
- DepthToScreen now correctly defaults to 1 as intended
- Fixed projectiles not being considered "cancelled" when other projectiles hit them
- Removed alpha from the criteria to determine if a sprite needs to be rendered. Normally it could be fixed (oversight in checking if destination is 255), but it could also cause trouble in the future if/when we add more blending types
- Fixes #2279 
- Fixes #2281 

Refactor:
- More functions of the type "lifeadd" will be skipped if the value is 0
- Fixed typo in shader error message
- Moved "noXguard" AssertSpecial flag resetting later into each frame. Previous placement was preventing IsAsserted trigger from seeing them properly
- Each projectile now carries the player number in its data


UltraCombo3000:
Not sure if there was an unintentional regression here, but this pull request has caused the latest nightly to generate a config.json file.

Edit: This is no longer true. No idea what happened.

PR #2303, opened 2025-02-04
feat: bgctrl3d & loopCount

Eiton:
New Extra properties for glTF models:
Node property: 
```
id: integer id used to identify the node in bgctrl3d
```
Animation property:
```
id: integer id used to identify the animation in bgctrl3d
loopCount: integer value. number of times the animation will loop. default is -1 (infinity)
enabled: string, set this to "0" or "false" to disable the animation.
```

Add new bgctrl3d to control glTF nodes and animations
type:
```
visible: change the visibility of the node with id specified in ctrlid
anim: enable / disable the animation with id specified in ctrlid
```
Example:
```
[BGctrldef test]
looptime = -1

[bgctrl3d test]
type = anim
time = 120
ctrlid = 1
value = 0

[bgctrl3d test]
type = visible
time = 120
ctrlid = 1
value = 0
```

### DIFF
--- a/src/stage.go
+++ b/src/stage.go
@@ -466,6 +466,8 @@ func (bg backGround) draw(pos [2]float32, drawscl, bgscl, stglscl float32,
 
 type bgCtrl struct {
 	bg           []*backGround
+	node         []*Node
+	anim         []*GLTFAnimation
 	currenttime  int32
 	starttime    int32
 	endtime      int32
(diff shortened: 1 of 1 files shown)


LasombraDemon:
Yay! One step closer to interactive 3D stuff!

I know I'm getting ahead of myself here, but we it could be useful giving these a SCTRLID, so we can target these via a future ModifyBGCtrl3D.

https://github.com/ikemen-engine/Ikemen-GO/wiki/Stage-features/#bgctrl_sctrlid

Issue #194, opened 2020-11-17
3ha's Bishamon's HCF + Kicks super is broken


LasombraDemon:
[BISHAMON.zip](https://github.com/Windblade-GR01/Ikemen-GO/files/5550522/BISHAMON.zip)

In Mugen 1.0, this works properly. In IKEMEN, it sends the opponent flying away too high and never "releases" it.

MUGEN:
![mugen000](https://user-images.githubusercontent.com/74574875/99325195-90678780-2854-11eb-8c5b-59280596df9a.png)

IKEMEN:
![ikemen001](https://user-images.githubusercontent.com/74574875/99325192-8fcef100-2854-11eb-8190-e6540fb09a29.png)
![ikemen002](https://user-images.githubusercontent.com/74574875/99325187-8e052d80-2854-11eb-8399-207779ac8adf.png)
![ikemen003](https://user-images.githubusercontent.com/74574875/99325197-9198b480-2854-11eb-9a4c-9028e59bbfbd.png)

K4thos:
I have very limited knowledge about mugen inner workings but what happens here doesn't make sense to me. CNS code: (with DisplayToClipboard addition)

```ini
[Statedef 3170]
type = U
movetype = H
physics = N
ctrl = 0
velset = 0, 0
sprpriority = 1

[State 3170, 1]
type = ChangeAnim2
trigger1 = Time = 0
value = 3190

[State 3170, 2]
type = LifeAdd
trigger1 = Time = 47
trigger2 = Time = 57
trigger3 = Time = 67
value = ifelse(Life > 80, -80, -(Life-1))

[State 3170, 3]
type = LifeAdd
trigger1 = Time = 117
value = -20

[State 3170, 3]
type = DisplayToClipboard
trigger1 = 1
text = "time = %d, Pos Y = %f, Vel Y = %f"
params = time, Pos Y, Vel Y

[State 3170, 4]
type = SelfState
trigger1 = time = 200
trigger1 = Pos Y > -20 &&  Vel Y > 0
value = 5100
```

Tested in mugen 1.1:
last frame of Kung Fu Man being custom stated:
![1](https://user-images.githubusercontent.com/8928195/99328628-c6b8fd00-287c-11eb-86dc-4bf744d4ca52.png)

and in the next frame he somehow enters state 5050:
![2](https://user-images.githubusercontent.com/8928195/99328674-dfc1ae00-287c-11eb-961a-321f096b83dc.png)

[...]

PR #2326, opened 2025-02-16
fix: Bindtime of 1 immediately after PosSets not unbinding in time

Jesuszilla:
Please let me know if this causes any other regressions (but I haven't found any related with Anakaris).

This fixes the following situation:
```ini
[State 20000, Bind]
type = PosSet
trigger1 = 1
x = floor(root, Pos X + root, Vel X)
y = 0
ignorehitpause = 1
[State 20000, BindToRoot]
type = BindToRoot
trigger1 = 1
time = 1
pos = 0, -RootDist Y
ignorehitpause = 1
```

In MUGEN 1.0 and 1.1, this works and binds the helper to the root. In IKEMEN-Go, the bind was lasting longer than it needed to, leading to no effect from the PosSet.

### DIFF
--- a/src/char.go
+++ b/src/char.go
@@ -8363,10 +8363,18 @@ func (c *Char) tick() {
 		if c.isBound() {
 			if bt := sys.playerID(c.bindToId); bt != nil && !bt.pause() {
 				c.bindTime -= 1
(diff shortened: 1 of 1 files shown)


NeatUnsou:
A regression has occurred.
https://github.com/ikemen-engine/Ikemen-GO/issues/194
I have confirmed that this fix has caused the problem to reoccur.

PR #2330, opened 2025-02-18
fix: TargetBind and Snap default facing behavior now handled differently from BindToParent/root default facing

Jesuszilla:
Default facing was broken in some instances for BindToParent/BindToRoot as discovered while working on Anakaris. Tested throw binds on Ultrarox_Clark, Warusaki3 Zangief, (WinMUGEN) and my WIP Anakaris (MUGEN 1.0 as of this writing), and helpers from Fusion's Zero (WinMUGEN) and my Anakaris (MUGEN 1.0).

### DIFF
--- a/src/bytecode.go
+++ b/src/bytecode.go
@@ -9801,7 +9801,7 @@ func (sc bindToParent) Run(c *Char, _ []int32) bool {
 	crun.bindPos[0] = x
 	crun.bindPos[1] = y
 	crun.bindPos[2] = z
-	crun.setBindToId(p)
+	crun.setBindToId(p, false)
 	crun.setBindTime(time)
 	return false
 }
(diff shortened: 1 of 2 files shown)


jenngra505:
This fixes the regression NeatUnsou pointed out in 3ha's Bishamon.

NeatUnsou:
I think the bug is not fixed

https://github.com/user-attachments/assets/caaf1bc5-99a9-4855-a397-b35237ee4f0b

NeatUnsou:
I forgot that I made it easier for bugs to occur when I tested it before.
Please test this with MUGEN1.1 and the nightly build.

[bishamon.zip](https://github.com/user-attachments/files/18871299/bishamon.zip)

PR #2344, opened 2025-03-01
feat:New screenpack parameters to load and render a glTF model

Eiton:
New screenpack parameters to load and render a glTF model:
[Files]
-model: path to a glTF model
[XXXBGdef]
-sceneNumber: scene number in the model to render
-modelOffset: x,y,z offset
-modelScale: x,y,z scale
-fov: fov of the view frustrum
-near: near of the view frustrum
-far: far of the view frustrum

New glTF Node extra parameter
-layerNumber: layer number of the node. Can be -1(stage model only), 0 or 1.

### DIFF
--- a/external/script/motif.lua
+++ b/external/script/motif.lua
@@ -33,6 +33,7 @@ local motif =
 		font_height = {},
 		glyphs = 'data/glyphs.sff', --Ikemen feature
 		module = '', --Ikemen feature
+		model = '', --Ikemen feature
 	},
 	ja_files = {}, --not used in Ikemen
 	music =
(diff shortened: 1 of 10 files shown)


K4thos:
@Eiton, looks like we have a merge conflict with your other merged commit

Eiton:
fixed

two4teezee:
This PR contains a lot of great features, but highlights some of the carelessness in which PRs can be approved on this project. Take `bgNew` from `script.go`, for instance.

'bgNew' already accepted three arguments, but in this PR a fourth argument was added. This new argument was added in the second position, which was previously occupied by something else, instead of including it in the last position as an optional argument,  thereby ensuring every module that uses `bgNew` is broken, causing a cascade of regressions.

Tagging @Eiton and @K4thos for awareness - if functions in `script.go` are modified/augmented (short of a major overhaul) we should be mindful of making these changes non-breaking. Should be pretty easy to catch when we review PRs.

PR #2484, opened 2025-06-09
fix: lifebar portraits offset, projection syntax

rakieldev:
Fixes:
- Fixes the offset issue with lifebar portraits in 16:9 resolutions
- Removed 'or', 'pe', and 'pe2' from the projection syntax


UrsoManso:
I was expecting https://github.com/ikemen-engine/Ikemen-GO/discussions/1870 to be in your next PR :(

rakieldev:
> I was expecting #1870 to be in your next PR :(

It’ll be in the next PR, this one was just for these small fixes

PR #2509, opened 2025-06-26
fix: font loading for AttachedChar

### LINKED ISSUE #2025: Attached Chars can't load fonts
### Describe the bug

Even though Attached Chars are essentially chars, they can't load fonts, Ikemen crashes if you try to do that.

### To Reproduce

1. Put an Attached Char in your stage.
2. Declare a valid font in the AC's .def.
3. Try to load the stage
4. To get the error log, load it via the command line
[ikemen-error-log.txt](https://github.com/user-attachments/files/17118547/ikemen-error-log.txt)

### Expected behavior

Load the fonts like the chars do.

### Screenshots / Video

_No response_

nightly 2024-09-23

### Operating system

Windows

_No response_

NeatUnsou:
・Fixed #2025 font loading for AttachedChar
・Fixed font loading from engine folder

### DIFF
--- a/src/char.go
+++ b/src/char.go
@@ -2810,9 +2810,7 @@ func (c *Char) load(def string) error {
 
 		if isZipDef {
 			if isEngineRootRelative {
-				// Example: kfm.zip/kfm.def, and kfm.def has `sound = data/common.snd`
-				// This should form kfm.zip/data/common.snd for SearchFile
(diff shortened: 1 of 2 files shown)


K4thos:
this exact problem was the reason why character fonts are preloaded at the start of the engine, if I remember correctly. I assume this fix also means that there won't be a crash when loading fonts for normal characters dynamically, so consider changing it to not preload them at all for all chars.

PR #2517, opened 2025-06-30
fix: Remove unnecessary .go file

SuperFromND:
Moves `framebufferSizeCallback` to system_glfw rather than a bespoke file, as created https://github.com/ikemen-engine/Ikemen-GO/pull/2511.

It's worth noting said commit appears to have been (at least partially) made by an AI of some sort judging by the commit author being a Google Labs bot. Personally speaking I'd be quite suspicious of its contributions.


SuperFromND:
Thought my PR wasn't working, turns out that no, scaling behavior was just broken before and after this commit. That behavior isn't affected by this, it seems.

PR #2524, opened 2025-07-09
fix(input): Make exclusive directions exclude opposing directions.

Alceris:
Without this, holding back and forward would trigger command = /F and command = /B at the same time, now it triggers neither of them as in MUGEN.


Alceris:
Hold up, I forgot half of this.

Issue #1551, opened 2023-12-09
Commands works different between mugen and ikemen


ManuLoo:
### Describe the bug

I seems that chars with cmd like this:

[Command]
name = "b"
command = a
time = 1

doesn't works correctly if we have something like this:

;Crouching B
[State -1]
type = ChangeState
value = 410
...
triggerall = command = "b"
triggerall = command = "holddown"
...

Mugen seems to interpret that you are pressing A and associate state 410 to A button, while Ikemen seems to interpret that you are pressing B and associate state 410 to B button.

### To Reproduce

[Benimaru XI UM.zip](https://github.com/ikemen-engine/Ikemen-GO/files/13625431/Benimaru.XI.UM.zip)

Download this character, and check that it hasnt strong kick and has two buttons for dodge in Ikemen, while in mugen works correctly

### Engine Version (or source code date)

0.99

### Operating system

Windows

potsmugen:
This is probably because of Ikemen's common commands. "B" will always be "B" because of them. If you're editing a character to switch the buttons around you should use the "Button Remapping" at the top of the char's command file.

ManuLoo:
Thanks for the reply!

Fix the character for me was more complicated than remap the buttons, because the specials movements works correctly; for example:

; Raikouken (low)
[Command]
name = "RaikoukenA"
command = ~D, F, D, F, x
time = 30

if i remap the X button, the special changes too. And, for the char that i upload, you don't have any button for the strong kick actually and the remap doesn't fix the problem.

The only way for me was read all the cmd and cns files and change all the single buttons declarations. In my first example, change all the single buttons like this 

[Command]
name = "a"
command = a
time = 1

and then change "b" to "a" wherever it are.

I fixed all the characters with this problem, but i wonder if it could be fixed in ikemen.

PR #2526, opened 2025-07-09
feat: Add PauseBuffer parameter to commands.

Alceris:
Commands with this parameter set to 0 will not be buffered during a pause's EndCMDBufferTime or HitPause.

This is mainly useful for implementing custom input systems without requiring the use of RunFirst and RunLast helpers.

### DIFF
--- a/data/common.cmd
+++ b/data/common.cmd
@@ -9,12 +9,14 @@ name = "TagShiftBack"
 command = d
 time = 1
 buffer.time = 1
+pausebuffer = 1
 
 [Command]
 name = "TagShiftFwd"
 command = w
 time = 1
 buffer.time = 1
+pausebuffer = 1
 
 [Command]
 name = "x"
--- a/src/compiler.go
+++ b/src/compiler.go
@@ -7484,6 +7484,7 @@ func (c *Compiler) Compile(pn int, def string, constants map[string]float32) (ma
 				if is.ReadI32("command.buffer.time", &i32) {
 					c.cmdl.DefaultBufferTime = Max(1, i32)
 				}
+				is.ReadBool("command.pausebuffer", &c.cmdl.DefaultPauseBuffer)
 			}
 		default:
 			// Read input commands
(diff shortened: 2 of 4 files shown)


potsmugen:
Pauses and hitpauses should probably be handled separately. A constant already exists to handle hitpauses, but I do also think it works better as a command parameter instead. Also, such parameters should probably also have default values (like `command.buffer.time`).

Maybe the EndCMDBufferTime override would also work as an AssertSpecial, too, since the normal behavior of commands is already to not buffer during pauses and EndCMDBufferTime itself is already an override. Like even if you set a command to have a pause buffer it's still reliant on that parameter.

Alceris:
I can't think of a single use case where you would want one but not the other. I mainly need this for my custom input system, which requires that I know exactly which frames that a direction and/or button is being held, and any buffering at all screws that up. And I really don't want to rewrite it to use InputTime because that would grossly overcomplicate it.

potsmugen:
It'll be just about the only place in the code where pause and hitpause are blended together, so when someone does come up with that use case it'll be the obvious place to refactor. Regardless, I think the syntax could be inverted and use `pausebuffer = 0` rather than `nopausebuffer = 1`.

Alceris:
Pause and HitPause are already blended together in terms of command buffering. It actually would be more complicated for little-to-negative benefit to make this parameter affect only one of them.

Alceris:
Any time I make a new feature 'do multiple things' it's because making it do only one of those things would require changing much more of the code and, at least in my mind, it'd be generally less useful.

potsmugen:
> Pause and HitPause are already blended together in terms of command buffering.

Only internally. I'm just saying if the refactor happens in the future it'll be a breaking change. You should also add the new parameter to `common.cmd`.

With this feature we can deprecate the `input.pauseonhitpause` constant.

Alceris:
There is a slight problem in that if a command in `common.cmd` doesn't have it set to 0 then that will interfere with characters that have the same command with that parameter set to 0. It's probably fine to set it to 0 for all of the ones that every character already has, at least.

potsmugen:
Oh, yeah, that makes sense. Then just add it to the tag inputs maybe. The functionality of the other common commands is not important per se and perhaps they're even harmful to begin with (see issue #1551).

Alceris:
Should it be 1 or 0 for the tag commands?

potsmugen:
1. Just so existing tag modules behave the same.

Alceris:
Done. I presume removing the other commands in there should be a separate PR.

potsmugen:
Yeah. That's a super easy fix to do even locally for the end user, so we have plenty of time to think about it. Common commands other than tag shift became obsolete when InputTime trigger was added.

PR #2528, opened 2025-07-11
fix: Remove bad common commands.

### LINKED ISSUE #1551: Commands works different between mugen and ikemen

Alceris:
They cause far more problems than they solve, and now with the existence of the InputTime trigger you can detect a button press without a command, solving the problems that these originally solved.

I presume the blank "recovery" command is there so that the common states referring to it don't crash the game if the character doesn't have the command, but is there any reason the game should crash from this to begin with?

Fixes #1551


potsmugen:
Even the "recovery" command doesn't need to be there since that was already required in Mugen. That is, a char that crashes Ikemen for not having it was already crashing Mugen. But that one should be more harmless than the others, and justified because of `training.zss`.

When I said remove them though I meant completely. Not just blanking the input.

potsmugen:
Also there's the original motive for having it:
https://github.com/ikemen-engine/Ikemen-GO/commit/f86abafdc92bb8e3a10174432e6e02960cc6d110

PR #2541, opened 2025-07-20
fix: Attachedchar input buffer is no longer written to the replay file

NeatUnsou:
・Attachedchar input buffer is no longer written to the replay file 
・Fixed a bug where musicvictory was not functioning 
・Fixes the regression with 3ha Bishamon after TargetBind #2332 #2330 #194


potsmugen:
We should probably comment that `bindTime` because I broke it recently thinking it needed the same change as the other one below it.

PR #2546, opened 2025-07-25
fix: Fixed HitBy/NotHitBy PlayerNo and PlayerID not working

NeatUnsou:
Fixed HitBy/NotHitBy PlayerNo and PlayerID not working

### DIFF
--- a/src/char.go
+++ b/src/char.go
@@ -7161,18 +7161,18 @@ func (c *Char) guardBreakEnabled() bool {
 func (c *Char) redLifeEnabled() bool {
 	return sys.lifebar.redlifebar
 	/*
-		switch sys.tmode[c.playerNo&1] {
-	    case TM_Single:
-	        return sys.cfg.Options.Single.RedLife
-	    case TM_Simul:
-	        return sys.cfg.Options.Simul.RedLife
-	    case TM_Tag:
-	        return sys.cfg.Options.Tag.RedLife
-	    case TM_Turns:
-	        return sys.cfg.Options.Turns.RedLife
-	    default:
-	        return false
-	    }
+			switch sys.tmode[c.playerNo&1] {
+		    case TM_Single:
+		        return sys.cfg.Options.Single.RedLife
+		    case TM_Simul:
+		        return sys.cfg.Options.Simul.RedLife
+		    case TM_Tag:
+		        return sys.cfg.Options.Tag.RedLife
+		    case TM_Turns:
+		        return sys.cfg.Options.Turns.RedLife
+		    default:
+		        return false
+		    }
 	*/
 }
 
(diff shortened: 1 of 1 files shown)


potsmugen:
Wonder when that broke. Maybe when I split the logic to add HitByAttr trigger.

It's been a while since I added the feature and to be honest I've forgotten some of the details, but this test case no longer works with the refactor:
```
[State -2, Test]
type = HitBy
trigger1 = 1
time = 1
slot = 0
attr = SCA, AA, AT, AP
playerno = 2
```
I think normally this would make the char hittable only by player 2.

It also looks like slots with stack will have higher priority than the rest.

Speaking of HitByAttr, the reason I made it ignore player number/ID parameters was the name of the trigger.

potsmugen:
It's been a while but, if it helps, all the stack parameter was meant to do was to evaluate the next slot before deciding if the hit should hit/miss. Essentially "continue" instead of "break". It allowed things that were impossible in Mugen like this (from the wiki when my memory was fresh, hehe):

```
Example, make a player vulnerable to Standing Attacks and Air Projectiles:

[State -2, Test]
type = hitby
trigger1 = 1
value = S, AA
time = 1
stack = 1

[State -2, Test]
type = hitby
trigger1 = 1
value2 = A, AP
time = 1
stack = 1
```

potsmugen:
Currently the first HitBy case works, but this one doesn't:

```
[State -2, Test]
type = NotHitBy
trigger1 = 1
time = 1
slot = 0
attr = SCA, AA, AT, AP
playerno = 1
```

Makes both players invincible.

NeatUnsou:
Thanks to you, able to fix several oversights.

Regarding the stack parameter, I believe it is at least functioning the same as the previous logic.
However, the sample code listed on the wiki does not work as intended. I'm not sure when this started, but in the implementation of bytecode.go, it seems that using the value parameter disables the stack, which feels natural.
The stack parameter currently seems to function by properly setting the attr and slot parameters.

```
[State -2, Test]
type = HitBy
trigger1 = 1
time = 1
slot = 0
attr = S, AA
stack=1

[State -2, Test]
type = HitBy
trigger1 = 1
time = 1
slot = 1
attr = A, AP
stack=1
```

potsmugen:
You're right. The new parameters only work with the new "slot" syntax. I forgot to update the example.

Edit: Won't dummyGetter maybe have a performance impact? I just realized. Especially when you use the trigger repeatedly like my clash module.

NeatUnsou:
That may be true. The function only needs the playerNo and id, so that's all we need.

potsmugen:
Dummy HitDef might have the same issue, but it's way smaller than Char at least. Isn't it the same scenario as with getter and playerno, where you only need the attr instead of a whole HitDef?

I had this random idea where instead of NotHitBy inverting the flags in bytecode, it could invert the logic in char.go, flipping true/false. That'd probably make expansions like playerno and future ones easier to implement. Actually now I kinda wish I had a chance to fix it, haha, but at least it's less work.

PR #2574, opened 2025-08-09
fix: Generate overdone commands automatically for specific commands

NeatUnsou:
・Generate overdone commands https://github.com/ikemen-engine/Ikemen-GO/issues/1280#issuecomment-2809708506 automatically for specific commands
This is a countermeasure for the issue where a command fails to register if directional inputs are performed more times than necessary to complete the command. #800

・Fixed #2572 an oversight with cellskip
・Fixed #1280
・Fixed #800

PR #2576, opened 2025-08-13
feat: Added the $N command

NeatUnsou:
・Added the $N command
In MUGEN 1.1, this is a fully functional command.
https://mugen-cheap.fandom.com/wiki/MUGEN's_Undocumented_Stuff#Unused_Command

・Changed all keys except the first key of extended inputs in autoGenerateCommand to $N. https://github.com/ikemen-engine/Ikemen-GO/pull/2574#issuecomment-3176864151
This allows for more flexible input handling, but more testing may be needed.

・Fixed a bug with cellfacing.
・Fixed an issue where char fightfx could not be loaded from the char folder.

### DIFF
--- a/external/script/start.lua
+++ b/external/script/start.lua
@@ -2111,9 +2111,6 @@ function start.f_selectScreen()
 
 	--cell drawList
 	local drawList = {}
-	local charfacing = motif.select_info['cell_' .. col .. '_' .. row .. '_facing'] or motif.select_info.portrait_facing
-	local randomfacing = motif.select_info['cell_' .. col .. '_' .. row .. '_facing'] or motif.select_info.cell_random_facing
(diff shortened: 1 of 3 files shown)


potsmugen:
In my testing, besides being undocumented,  Mugen's N symbol is too inconsistent to be of use. When I added N to Ikemen, it was to represent neutral direction, which is a common fighting game concept. Hence "any neutral" ($N) meaning the same as just neutral.

NeatUnsou:
There is very little information available regarding the N command, but it certainly exists.
http://minormugen.blog.fc2.com/blog-entry-78.html
The fact that it exists and functions in MUGEN is, I think, sufficient reason to port it to IKEMEN GO.

UrsoManso:
Still faulty.

https://github.com/user-attachments/assets/7718cfe2-e06d-44f6-9735-c632043ee5c9

In this case doing QCB+K > QCBQCF+K only works, once again, by delaying it a little.

PR #2580, opened 2025-08-15
fix: -speed and -speedtest command line arguments

### LINKED ISSUE #2385: Command line speedtest not working
### Describe the bug

Running Ikemen with `speed`, `speedtest` or `stresstest` arguments (maybe others) causes a Lua crash. Maybe it was the game speed refactor?

### To Reproduce

1. Run Ikemen from command line with those arguments

### Expected behavior

Same as previous builds.

### Screenshots / Video

_No response_

20/03/2025

### Operating system

Windows

_No response_

arroyoquiel:
Fixed #2385

Previously, using the command line arguments `-speed <speed>` and `-speedtest` crashed the game, but with these changes the game will no longer crash and will properly change the speed of the match.

Incredibly useful for bulk character testing, such as plotting win rates.

### DIFF
--- a/external/script/main.lua
+++ b/external/script/main.lua
@@ -351,11 +351,14 @@ end
 if main.flags['-ailevel'] ~= nil then
 	modifyGameOption('Options.Difficulty', math.max(1, math.min(tonumber(main.flags['-ailevel']), 8)))
 end
-if main.flags['-speed'] ~= nil and tonumber(main.flags['-speed']) > 0 then
-	setGameSpeed(tonumber(main.flags['-speed']) * gameOption('Config.Framerate') / 100)
+if main.flags['-speed'] ~= nil then
+	local speed_val = tonumber(main.flags['-speed'])
+	if speed_val ~= nil and speed_val >= -9 and speed_val <= 9 then
+		modifyGameOption('Options.GameSpeed', speed_val)
+	end
 end
 if main.flags['-speedtest'] ~= nil then
-	setGameSpeed(100 * gameOption('Config.Framerate'))
+	modifyGameOption('Options.GameSpeedMultiplier', 100)
 end
 if main.flags['-nosound'] ~= nil then
 	modifyGameOption('Sound.MasterVolume', 0)
--- a/external/script/main.lua
+++ b/external/script/main.lua
@@ -352,13 +352,25 @@ if main.flags['-ailevel'] ~= nil then
 	modifyGameOption('Options.Difficulty', math.max(1, math.min(tonumber(main.flags['-ailevel']), 8)))
 end
 if main.flags['-speed'] ~= nil then
-	local speed_val = tonumber(main.flags['-speed'])
-	if speed_val ~= nil and speed_val >= -9 and speed_val <= 9 then
-		modifyGameOption('Options.GameSpeed', speed_val)
+	local speed_input = tonumber(main.flags['-speed'])
+	if speed_input ~= nil and speed_input >= -9 and speed_input <= 9 then
+		local target_game_speed
+		if speed_input > 0 then
+			target_game_speed = speed_input
+		elseif speed_input == 0 then
+			target_game_speed = 0
+		else
+			if speed_input >= -1 then
+				target_game_speed = speed_input * 5
+			else
+				target_game_speed = speed_input * 0.5 - 4.5
+			end
+		end
+		setGameSpeed(target_game_speed)
 	end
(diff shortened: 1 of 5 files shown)


potsmugen:
This is better than not having it and should be merged, but maybe in the future -speed should be like Mugen and use the same speed settings as the ingame options.

arroyoquiel:
> This is better than not having it and should be merged, but maybe in the future -speed should be like Mugen and use the same speed settings as the ingame options.

You're completely right, after trying with MUGEN 1.1, the `-speed` argument uses an integer range [-9,9] just like the `GameSpeed` key in `data/mugen.cfg`.

Or it can also be edited to have a percentage limit, and reflect the [wiki documentation](https://github.com/ikemen-engine/Ikemen-GO/wiki/Miscellaneous-info#command-line-arguments) `-speed <speed>: Changes game speed setting to <speed> (10%-200%)`

It would be great to have this merged while I find a solution for the `-stresstest <value>` command which is not limited to a percentage.

arroyoquiel:
While testing, I also encountered another bug with parameters, the command line parser was misinterpreting arguments, causing crashes.

1. `Ikemen_GO.exe -p1 kfm -p2 kfm -p1.ai 8 -p2.ai 8 -speedtest`
Worked, no conflicts
2. `Ikemen_GO.exe -p1 kfm -p2 kfm -speedtest -p1.ai 8 -p2.ai 8`
Crashed because the parser, expecting a value for -speedtest and took -p1.ai as the value, leaving 8 to be misinterpreted as a file `8/8.def`
3. `Ikemen_GO.exe -p1 kfm -p2 kfm -speedtest 1 -p1.ai 8 -p2.ai 8` 
Crashed because 1 was incorrectly as 1/1.def
4. `Ikemen_GO.exe kfm kfm -speedtest -p1.ai 8 -p2.ai 8`
Similar issue to the second test

So I made a commit to fix this, the only thing left would be to fix `-stresstest`.

And also...

> This is better than not having it and should be merged, but maybe in the future -speed should be like Mugen and use the same speed settings as the ingame options.

The `-speed` parameter was limited to [-9,9], minimum speed of 25% and a max of 175%, anything outside that range will be ignored like in MUGEN.

potsmugen:
I thought -speedtest didn't take any arguments? Good job either way.

arroyoquiel:
> I thought -speedtest didn't take any arguments? Good job either way.

Thank you! And it definitely doesn't 😅, I wanted to reproduce how MUGEN handles these command line parameters, so if it doesn't crash in MUGEN, it shouldn't crash in Ikemen. Useful for other third-party tools like FF or my ikemen-manager-go tool that I plan to release in the near future

potsmugen:
In the future we could make bad arguments print an error and not load. A lot of software ignores bad arguments and leaves you scratching your head about what didn't work.

arroyoquiel:
I think this can be merged for now while I try to get frame skip working, I've been trying too hard for too long.

K4thos:
I don't see the point of adding GameSpeedMultiplier as a new option in the config.ini file when GameSpeed already exists. It makes more sense to add the missing setGameSpeed function to script.go and set the gameSpeed value from the system struct there (currently unused, which was a mistake made when adding ini support). Then, use that value as s.gameSpeed instead of s.cfg.Options.GameSpeedMultiplier.

arroyoquiel:
> Then, use that value as s.gameSpeed instead of s.cfg.Options.GameSpeedMultiplier.

Thanks for the feedback! I'll make the update ASAP, unfortunately my computer's PSU exploded, so I'll get on it when my PC gets fixed 😬

potsmugen:
Wow we've just had that `gameSpeed` var there gathering dust who knows how long. I think the rest of the codebase more often accesses the options directly instead of duplicating them in some variable. As seen in this very same line you updated:

`spd := ((60 + s.gameSpeed*5) / float32(s.cfg.Config.Framerate)) * s.accel`

Maybe drop `gameSpeed` var for good?

arroyoquiel:
@potsmugen I think this is finally it 😮

potsmugen:
Final nitpick I guess in that this feature used to be way faster in version 0.99. Used to finish a round in a couple of seconds. I guess we need to add another 0 to the right of speed, heh.
Here's the old formula:
```
if main.flags['-speedtest'] ~= nil then
	setGameSpeed(100 * config.Framerate)
end
```
I think in this old logic the normal game speed was 60.

Issue #2564, opened 2025-08-05
XYTruncate creating Mugen inaccuracy


potsmugen:
### Describe the bug

If you enable XYTruncate, RGB sprites won't move as smoothly as in Mugen, creating a scenario where you're damned if you do and damned if you don't with that toggle.

### To Reproduce

The stage from the videos:
https://www.mediafire.com/file/fytbblf2y7banhb/TimelessChamber.zip/file

### Expected behavior

I think the parameter should either default to disabled, or (maybe better) we should truncate 8bit but not 32bit and skip the toggle altogether.

### Screenshots / Video

Mugen 1.1:

https://github.com/user-attachments/assets/dc0ee7b7-3c69-4bf5-b752-57a84b8ad693

Ikemen:

https://github.com/user-attachments/assets/b6cfbae2-51c2-4dea-815d-f6d134c0640c

(Ignore the moving KFM because it's a leftover test, hehe)

### Engine Version (or source code date)

Nightly

### Operating system

None

Kasasagi77:
The problems are definitely not limited to 32-bit sprites.

XyTruncate also explodes the number and visibility of sub-pixel artifacts and/or jitter for any sprites/animations that need to be aligned to a local-pixel, in comparison to both Mugen 1.1 and any other previous version of Ikemen.

Historically, Ikemen's rendering always had more of these types of artifacts than Mugen 1.1, but this single change escalates the problem immensely.

Just a few examples based on my stage edits (but I can provide many more):

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/028b1ed2-c5e0-4ef0-b555-5d78df856448" />

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/2fa566f5-49f6-45bf-8505-ae7ede091bfb" />

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/c8580ef5-eb28-4aee-9f07-d0f465cce182" />

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/1a640b86-f927-45b2-86fa-24742cc3480a" />

The above examples are for a build from 2025-08-11, with the latest fixes for XyTruncate, and for the 1920x1080 resolution.

You can download the test stages from these examples here: https://drive.proton.me/urls/N1RA3H2C5G#6ehwyEb4X2jh

---

I have a strong opinion that XyTruncate should be unconditionally turned off by default.

[...]

PR #2582, opened 2025-08-16
fix: XY-Truncate option is now turned off by default

Kasasagi77:
Related to: #2564

Until some other consensus is reached about the ultimate fate of XY-Truncate I propose at least turning it off by default, for it to not cause any more problems for the nightly users.


Kasasagi77:
@potsmugen
@SuperFromND

SuperFromND:
I'm okay with this!

PR #2604, opened 2025-08-29
feat(motif): added Mugen 1.1 select screen cursor tween feature

rakieldev:
Feat:
- Adds the Mugen 1.1 cursor tweening feature as a configurable toggle in system.def [Select Info] with 3 new parameters:

- ``cursortween = 1``
Enables cursor tweening (default is 0)

- ``cursortween.speed = x, y``
Sets the cursor speed when moving between cells

- ``cursortween.wrap.speed = x, y``
Sets the speed to use when the cursor wraps around (wrapping = 1), If not declared or set to 0, it defaults the x speed to cursortween.speed multiplied by the number of columns, and the y speed by the number of rows


Kasasagi77:
Thank you for this!

If this is what I think it is, then it fixes #1377, an issue I reported  a long time ago, which was promptly demoted to a "discussion" for not beign "important" enough, despite clearly being a missing Mugen 1.1 feature :stuck_out_tongue:

rakieldev:
> Thank you for this!
> 
> If this is what I think it is, then it fixes #1377, an issue I reported a long time ago, which was promptly demoted to a "discussion" for not beign "important" enough, despite clearly being a missing Mugen 1.1 feature 😛

Yeah, that’s the one. But instead of being hardcoded, you can now enable, disable, and customize it however you want

PR #2611, opened 2025-09-01
feat(motif): menu.item.bg and select cursor.reset

wily-coyote:
### Menus
- Added `menu.item.bg` and `menu.item.active.bg`, a way to give every itemname a common background
  - This is different from applying a background to a specific itemname (`menu.bg.<itemname>.{spr,anim}`), which is also rendered separately

### Select screen
- Added `p<n>.cursor.reset`, which resets the cursor animation when the user moves to a new cell

### Fixes
- Removed duplicate `motif.lua` entries


K4thos:
All parameters should be in system.base.def. The same goes for your previous Lua-related commits. Also, I’m not sure how deep you plan to go with the new customization features, but keep in mind that when it comes to motifs, some may be better suited for external modules.

Issue #2600, opened 2025-08-27
Extreme performance degradation after repeatedly switching BGM/music tracks


Kasasagi77:
### Describe the bug

After switching the BGM/music track rapidly and repeatedly several times (by e.g. entering and exiting different types of screens an/or the fight) the performance drops by orders of magnitude: sound stutters, framerate drops significantly.

The effect is usually temporary, it can resolve itself after e.g. 30 seconds or after entering a fight, but it can also last longer when staying e.g. on the select screen.

Reproducible for different screenpacks (e.g. for MUGEN1 Screenpack Remix by Phantom.of.the.Server), so most probably is not dependent on any particular screenpack. Tested on a roster with a lot of characters (maybe like a 100?); I do not know if this is a requirement.

My hardware is decidedly not the fastest, but I had no performance problems with Ikemen Go Nightly previously.

The problem is definitely not that old, months at most.

### To Reproduce

1. In the title/menu screen: choose a game mode (any will do, e.g. "arcade") and enter the select screen (or alternatively enter the fight directly if e.g. the "randomtest" mode was selected).
2. Go back to the title/menu screen.
3. Repeat steps 1 and 2 five times or more (performace gets progressively worse with each repeat).
4. Observe sound stuttering and a significant framerate drop.

This is 100% reproducible on my setup.

The screenpack needs to have music tracks on the tested screens.

In fact, you can alternatively reproduce this also by switching between the fight and the select screen rapidly without ever entering title/menu screen.

### Expected behavior

No performance drop when switching the BGM/music tracks.

### Screenshots / Video

Not currently.

### Engine Version (or source code date)

2025-08-19

### Operating system

Linux

### Extra context or search terms

Since the problem resolves itself in time, it may be garbage collector related? Maybe something repeatedly allocates a lot of memory when switching to a different screen type? This is just a wild guess.

potsmugen:
Just a feeling but try disabling the screenpack music.

Kasasagi77:
> Just a feeling but try disabling the screenpack music.

Aaand.. we have a winner 😄 

After disabling the screenpack music:

1. switching between title/menu and select screens (both have no music) => no problem,
2. switching between title/menu and fight (where the stages have music) => **problem still exists**.

So the problem is reproduced when switching music tracks.

Kasasagi77:
I was wondering who to notify about this issue, and looking back a the previos PRs, I think maybe @Jesuszilla ?

[...]

PR #2630, opened 2025-09-06
feat: Made BGM RAM Swap a toggleable feature; fix: possible deadlock in SwapSeeker; build fixes

Jesuszilla:
* Fixes a possible deadlock scenario that could occur with lock acquisition in SwapSeeker.Swap()
* If that doesn't fix it, then disabling the RAM swap entirely should (at the cost of skipping, nothing we can do about that at this point from what I can tell, otherwise the main thread would be stuck waiting on the BGM to begin to buffer & play)

Meant as a resolution for #2600 

Build fixes:
* Added arenas to the environment variables for all builds (required for rollback).
* Added macOS ARM target and made the previous macOS target specifically target x64 (macs would otherwise default to their own architecture which could make cross-distribution problematic for users still stuck on Intel)
* Updated app bundle run script to handle new macOS executable names.

As mentioned before, the pipeline probably needs to be changed to accommodate the appbundle now, not sure what needs to be done there but I will say the .app is a specialized folder in macOS, so it will appear as such on other systems.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,6 +43,14 @@ Ikemen_GO_LinuxARM: ${srcFiles}
 # MacOS x64 target
 Ikemen_GO_MacOS: ${srcFiles}
 	cd ./build && bash ./build.sh MacOS
+	$(MAKE) clean_appbundle
+	$(MAKE) appbundle BINNAME=Ikemen_GO_MacOS
+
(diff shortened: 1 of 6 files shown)


Kasasagi77:
Thank you again for making the BGM RAM swapping optonal.

PR #2638, opened 2025-09-08
build: last attempt at building macOS app bundle through pipeline

Jesuszilla:


Jesuszilla:
This probably won't work due to the cd deploy

PR #2672, opened 2025-09-20
fix: rollback desync, SuperPause effects, explod regressions; refactoring

### LINKED ISSUE #1274: Simultaneous super pauses
### Describe the bug

If two characters call a super pause on the same frame, the super pause effect will only show for one of them. In Mugen it shows for both.

However, ideally, you'd have the pause happen for one player, and only once that was over would it happen for the second player. This is how it works in the few commercial games I tested.

### To Reproduce

1. Paste this into KFM's negative states:
```
[State -2, Super]
type = SuperPause
trigger1 = !(gametime % 10)
trigger1 = roundstate = 2
time = 5
pos = 25, -57
anim = 100
sound = 20, 0
poweradd = -1000
```

### Expected behavior

Same as Mugen 1.1, or ideally as commercial games.

### Screenshots / Video

Ikemen:
ikemen004~

Mugen 1.1:
mugen064

09/06/2023

### Operating system

Windows

_No response_

### LINKED ISSUE #1719: Custom Common FX palette issue
### Describe the bug

Custom hitsparks using the first palette index have issues displaying properly when character def loads its first palette. Other sparks using indexes after the first load the correct palettes regardless of character def palette.

### To Reproduce

1. Download SF6 Common FX. [Download](https://www.mediafire.com/file/lvmanzkc86f4g9i/sf6fx.zip/file) Add to your save/config.json.
```
  ],
  "CommonFx": [
    "data/SF6.def"
  ],
```
2. Add this code to an attack with more than one palette (light punch stateno 200 if available).

```
[State 200, hit spark]
type = explod
trigger1 = movehit = 1
anim = drive0
space = stage
facing = facing
pos = ifelse(facing = 1, 1, -1) * 55 + pos x + camerapos x , -87 + pos y
scale = 0.38, 0.38
sprpriority = 2
ownpal = 1
ignorehitpause = 1
persistent = 0
```
3. Run character in both pal1 and palx (x being any other palette than 1). 
4. Error is pal1 displays this spark properly. Palx displays the sprite in a wrong index. I think it's displaying it in index 0,1.
5. You'll notice that if you change the explod anim from drive0 to drive1 or drive2, these sprites would display properly when using character def palx. Seems this issue is only the first custom fx sprites index and for character def palx.
6. The workaround fix is just avoid using the first pal index on your custom fx. 

### Expected behavior

I would expect the customfx to load the first indexed palette for these sprites using any character def palette.

### Screenshots / Video

https://github.com/ikemen-engine/Ikemen-GO/assets/62481815/1d0f3b00-5f28-4d5a-ab39-eb1641438fcf

latest nightly as of 4/1/24

### Operating system

Windows

_No response_

### LINKED ISSUE #2664: Divinewolf's chars cloning
### Describe the bug

While testing some characters on training, everytime divinewolf's characters gets knocked down and hits the ground, they create a clone of themselves but then disappears.

### To Reproduce

1. Download character from this website. https://divinewolf.trinitymugen.net/chars/ pick any characters except the "DW" ones.
2. Add them to the latest version and use them in training mode. Knock them to the ground
3. See error

### Expected behavior

Divinewolf's characters should create clones everytime they hit the ground

### Screenshots / Video

_No response_

Nightly 16-09-25

### Operating system

Windows

_No response_

potsmugen:
Fix:
- Fixed a typo that apparently made moves that relied on GameTime often cause desync
- Changing into the enemy's sprites with SpritePlayerNo now automatically corrects localcoord scale
- Made GameMakeAnim compatible with explod refactor
- Fixed explod PalFX being applied twice in some scenarios
- The SuperPause effects are now explods instead of objects that have to be tracked by system.go. This simplifies the code as well as allows multiple effects when multiple pauses happen simultaneously
- If multiple players initiate a SuperPause, they now all ignore the "darken" parameter
- SuperPause p2defmul can now stack if multiple players trigger it in the same frame
- Fixed "workingchar" getting a full clone in save states, when it's just a pointer to the character which code is currently running
- All common FX prefixes now ignore palette remaps. Previously they were only skipped if the FX prefix was "F"
- A net connection is now only initialized after both peers complete a handshake. This makes our IsConnected() check work as intended, so if a client tries to join before the host is hosting, it will automatically retry until a session exists
- Patched menu confirmation sounds into "host" and "join" options
- Fixes #1274
- Fixes #1719
- Fixes #2664

Feat:
- KeepSpritesOnReload debug option. Enabling this option makes character SFF files not be reloaded when shift F4 is used, making debugging faster

Refactor:
- Refactored anim/sprite getting functions. Now when we "get" an animation we can also get a SFF with it
- GameMakeAnim no longer supports "anim" syntax. Only "value", like Mugen
- Some data used values or pointers arbitrarily. More of them now use pointers for consistency and rollback performance, such as projectiles and explods
- Removed more unused rollback functions
- Extracted some rollback code blocks into dedicated functions

### DIFF
--- a/external/script/main.lua
+++ b/external/script/main.lua
@@ -2794,6 +2794,7 @@ main.t_itemname = {
 	end,
 	--SERVER CONNECT
 	['serverconnect'] = function(t, item)
+		sndPlay(motif.files.snd_data, motif[main.group].cursor_done_snd[1], motif[main.group].cursor_done_snd[2]) -- Needs manual sndPlay due to special menu behavior
 		if main.f_connect(gameOption('Netplay.IP.' .. t[item].displayname), main.f_extractText(motif.title_info.connecting_join_text, t[item].displayname, gameOption('Netplay.IP.' .. t[item].displayname))) then
 			synchronize()
 			math.randomseed(sszRandom())
@@ -2807,6 +2808,7 @@ main.t_itemname = {
 	end,
 	--SERVER HOST
 	['serverhost'] = function(t, item)
+		sndPlay(motif.files.snd_data, motif[main.group].cursor_done_snd[1], motif[main.group].cursor_done_snd[2]) -- Needs manual sndPlay due to special menu behavior
 		if main.f_connect("", main.f_extractText(motif.title_info.connecting_host_text, gameOption('Netplay.ListenPort'))) then
 			synchronize()
 			math.randomseed(sszRandom())
--- a/src/anim.go
+++ b/src/anim.go
@@ -173,10 +173,6 @@ type Animation struct {
 	start_scale                [2]float32
 }
 
-func (a *Animation) isBlank() bool {
-	return a.scale_x == 0 || a.scale_y == 0 || a.spr == nil || a.spr.isBlank()
-}
-
 func newAnimation(sff *Sff, pal *PaletteList) *Animation {
 	return &Animation{
 		sff:         sff,
@@ -345,6 +341,19 @@ func (a *Animation) Reset() {
 	a.curelemtime, a.curtime = 0, 0
 	a.newframe, a.loopend = true, false
 	a.spr = nil
+}
+
+func (a *Animation) isBlank() bool {
+	return a.scale_x == 0 || a.scale_y == 0 || a.spr == nil || a.spr.isBlank()
+}
+
+func (a *Animation) isCommonFX() bool {
+	for _, fx := range sys.ffx {
+		if fx.fsff == a.sff {
+			return true
+		}
+	}
+	return false
 }
 
 func (a *Animation) AnimTime() int32 {
--- a/src/bytecode.go
+++ b/src/bytecode.go
@@ -9367,17 +9367,20 @@ func (sc superPause) Run(c *Char, _ []int32) bool {
 		return false
 	}
 
-	sys.superpmap.remap = nil
-	sys.superpos = [2]float32{crun.pos[0] * crun.localscl, crun.pos[1] * crun.localscl}
-	sys.superscale = [2]float32{crun.facing, 1}
-
 	var t, mt int32 = 30, 0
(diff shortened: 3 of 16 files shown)


softmgr:
@potsmugen  [Bug Report] When anim = -1, `SuperPause` still darkens the screen.

It is known that in M.U.G.E.N 1.1, when `anim = -1`, the game is only paused according to the value of time, but the screen does not darken.
However, in Ikemen-Go, the screen darkens regardless of how anim is set.
For example:
```
[State 0, SuperPause]
type = SuperPause
trigger1 = animelem = 3
time = 35
anim = -1
pos = 0,0
```

PS: In older versions, we could determine anim = -1 by checking s.superanim == null.
```
func (s *System) draw(x, y, scl float32) {

	s.brightness = 0x100 >> uint(Btoi(s.supertime > 0 && s.superdarken && s.superanim != nil))  //s.brightness = 0x100 >> uint(Btoi(s.supertime > 0 && s.superdarken))

}

```
In the new master branch, this variable was removed.
As a result, in the above M.U.G.E.N code, it is no longer possible to use s.superanim, which causes the screen to always darken.
This makes the interpretation of traditional M.U.G.E.N code inconsistent with the original M.U.G.E.N behavior.

potsmugen:
Are you sure? Darken seems to default to true in Mugen regardless of animation, as documented.

softmgr:
> Are you sure? Darken seems to default to true in Mugen regardless of animation, as documented.

I retested it with M.U.G.E.N 1.1, and it turns out I was mistaken: whether the screen darkens or not is unrelated to anim = -1.
The screen darkening is controlled by `darken = 1`, and to prevent the screen from darkening you should use `darken = 0`.

potsmugen:
Yeah. So it looks like Ikemen was behaving the wrong way and it was accidentally fixed.

Issue #168, opened 2020-08-23
[Bug] RemapPal does not apply to helpers under certain conditions.


jenngra505:
I noticed this particular bug while testing my own creation, Numbuh 3, on IKEMEN GO.
I made a benchmark to test when this feature works.
https://1drv.ms/u/s!AqN1Ini5-I2Ig-gmlu7gjfFljqY7ng?e=vDFhZs
Update: Ownpal being set to 0 is ignored in Mugen default but works properly on IKEMEN.
Palette on helper still doesn't change when ownpal is set to 1. Benchmark updated.

K4thos:
this problem was investigated and solution was found on discord.

Relevant posts:
> K4thos — 18.11.2020
the problem is in sff file. I will see what's up but after reimporting sprites in fighter factory studio test case works fine:
https://cdn.discordapp.com/attachments/233363722934943744/778604287806537778/Ikemen_GO_2020-11-18_13-45-51.mp4

> Jenngra505 — 18.11.2020
I had a feeling, because identical code in an unfinished character of mine worked on IKEMEN but Numbuh 3 still didn't work.

>Jenngra505 — 29.01.2021
I think I realized why Numbuh 3's sff causes IKEMEN GO problems, when I started developing her, I imported a 9000px tall spritesheet and cropped the individual sprites out within Fighter Factory.
https://sta.sh/0202g6wp3kys

jenngra505:
I added a working version of the sff to the RemapPal Test, if someone wants to compare bytecodes in a hex editor go ahead.

jenngra505:
I created minimalist versions of the Broken and Working sff files and noticed a 2-kilobyte size difference. I uploaded these to the benchmark for further investigation.

jenngra505:
Lunatic404 investigated the sff on Discord and discovered that the broken version crammed multiple palettes into the same byte offset while the working version didn't. Oddly, Fighter Factory never notices this. I'll release a patch for Numbuh 3 that makes her graphics work on IKEMEN after I release my first IKEMEN-exclusive character, now that I know why it didn't work.

PR #2677, opened 2025-09-21
docs: Added comment about Mugen RemapPal inconsistency.

jenngra505:
I added a comment in the RemapPal section of the bytecode to point out an unfixable inconsistency with Mugen based on what was discovered in #168


potsmugen:
Is that commented related to RemapPal itself or to the SFF format?

PR #2702, opened 2025-09-26
fix: game states, helpers, options, explods, NotHitBy; misc refactoring

### LINKED ISSUE #2431: Ingame options bilinear filtering writes to WindowScaleMode
### Describe the bug

Changing bilinear filtering ingame will change `WindowScaleMode` in `config.ini`.

[Posting on behalf of Horse](https://discord.com/channels/233345562261323776/233363722934943744/1363087023489155285)

### To Reproduce

1. Go to Video and toggle bilinear filtering
2. Check changes in config.ini

### Expected behavior

_No response_

### Screenshots / Video

Image

2025/04/16

### Operating system

Windows

_No response_

### LINKED ISSUE #2614: Game options bool/int inconsistency
### Describe the bug

I noticed that when you first get Ikemen, the default config.ini uses int bools (0/1), but when you change an option it saves it as a true bool (true/false).

### To Reproduce

1. Go into ingame options
2. Change for instance Dizzy to "Yes"
3. config.ini will change from "0" to "true"

### Expected behavior

Only 0/1 I suppose? Like mugen.cfg.

### Screenshots / Video

_No response_

31/08/2025

### Operating system

Windows

_No response_

potsmugen:
Feat:
- GameSpeedStep option. Unhardcodes the value of each game speed step
- Ctrl Alt number now also works on the (first) attached char

Fix:
- Rollback states no longer clone the "debug working char"
- Helpers with preserve parameter will be "rewinded" when F4 is used instead of ignoring it
- If the second value of RemapPal dest is not defined, it will stay at default (0) as in Mugen
- Fixed explod drawing offset when z-axis is used in 16:9
- Boolean options now write to config.ini as 1/0 instead of true/false as intended
- Fixed characters pushing each other incorrectly along the z-axis if their localcoords were different
- Fix NotHitBy evaluation to consider both player number/ID and attack attributes correctly
- Fixed "stack" slots making non stack ones be ignored

Refactor:
- Improved GameVar(slowtime) implementation. System functions and trigger version use the same source of truth
- A few compiler restrictions that previously depended on the character's engine version now depend on the state file format (ZSS or CNS) instead
- WindowScaleMode ingame settings renamed from "disabled" and "enabled" to "nearest" and "bilinear". Ingame option name now same as internal name, since we no longer need to clarify what it does
- When config.ini is saved it will normalize all bools to 1/0
- Removed explod layer lists from the code. The main explod slice is now used for all purposes, making the code less complex
- Adjusted the way explods are removed. A new explod will never occupy the old slot of a removed explod, making their game and drawing logic more predictable
- Removed char "draw order" list since we were already not really using it
- Adjusted collision detection to first collect all Clsn proxies instead of using recursion
- IsHelper trigger and Helper redirection now scan the entire player list in order, making them more accurate and keeping indexes defragmented
- PlayerID now skips helpers under DestroySelf effect, as in Mugen. Index, NumPlayer and PlayerIndex now also do the same
- Simplified (Not)HitBy logic a little. Now both run the same code, with NotHitBy merely setting the "not" flag
- Merged some projectile initialization steps that were run together, making the logic easier to follow

Fixes #2431 
Fixes #2614


NeatUnsou:
It seems there has been a regression in NotHitBy.
In NotHitBy specified only with value = SCA, it should grant invincibility against all attributes, but currently, it doesn't function.

```
[state ]
type = NotHitBy
trigger1 = 1
value = SCA
```

potsmugen:
You're right. With my refactor I'm trying to make it easier to expand if we ever want to add more parameters. But it looks like the logic can't be as simple as I wished. Think I've fixed it now but need to test more scenarios.

PR #2752, opened 2025-10-08
fix: system.base.def and motif.lua formating

rakieldev:
Minor corrections


UrsoManso:
Under Victory Screen section in system.base.def it's listed the old parameter p2.pal instead of the new one p2.applypal. Also some of the bg elements have a underscore such as p2.palmenu.bg_offset, p2.palmenu.bg_facing, and p2.palmenu.bg_scale.

Issue #2771, opened 2025-10-11
Inconsistent character swapping pattern during random select.


UrsoManso:
### Describe the bug

Character swapping pattern during random select in Ikemen seems to not be as consistent as MUGEN's.

### To Reproduce

Set up at least 2 different characters and place the cursor on random select.

### Expected behavior

For the random select quickly swap between characters.

### Screenshots / Video

MUGEN 1.1

https://github.com/user-attachments/assets/1778a482-6c2d-47d9-af2c-b88741e9bb30

Ikemen GO

https://github.com/user-attachments/assets/254bfbba-2355-4e25-8d80-abce40df62af

### Engine Version (or source code date)

Nightly

### Operating system

Windows

potsmugen:
Looks like Mugen is real random while Ikemen only shows each char once per random cycle. Maybe.

formula1:
Could be that Ikemen is true random while mugen removes the previous character from being chosen.

True random might select a character more than once which causes a character to be selected multiple frames in a row
```typescript
function chooseRandom(allAvailableCharacters: Array<Character>){
  return function selectNext(){
    const index = Math.floor(Math.random() * allAvailableCharacters.length);
    return allAvailableCharacters[index];
  };
}
```

Meanwhile, mugen seems not to select a previously selected character when choosing next random
I think something like below
```typescript
function chooseRandomFromUnchosen(allAvailableCharacters: Array<Character>){
  let unchosenCharacters = [...allAvailableCharacters];
  return function selectNext(){
    if(unchosenCharacters.length === 0){
      unchosenCharacters = [...allAvailableCharacters];
    }
    const index = Math.floor(Math.random() * unchosenCharacters.length);
    const chosen = unchosenCharacters[index];

    unchosenCharacters[index] = unchosenCharacters[unchosenCharacters.length - 1];
    unchosenCharacters.pop();

    return chosen;
  };
}
```

Personally, I think the second might be better to avoid a character being selected more than once like the second video seems to do

---

Is this the correct code?
https://github.com/ikemen-engine/Ikemen-GO/blob/develop/src/system.go#L3934
```go
func (s *Select) AddSelectedChar(tn, cn, pl int) bool {
	m, n := 0, s.GetCharNo(cn)
	if len(s.charlist) == 0 || len(s.charlist[n].def) == 0 {
		return false
	}
	for s.charlist[n].def == "randomselect" || len(s.charlist[n].def) == 0 {
		m++
[...]

PR #2774, opened 2025-10-11
feat: Avoid repeats for random select 

formula1:
Related to #2771 

# Help Needed
- @potsmugen mentioned that there should be an option in the `config.ini`, I don't know where that is. I saved the old code for that though
- Looks like most of the random select is handled in Lua so I'm not sure if the Go code is necessary. We can remove it since statistically choosing randomselect that many times is unlikely.

# Changes
## Go
- Creates a list of valid characters to choose from when resetting
- When randomselect is chosen for a character, choose from the valid characters

## Lua
- Created unchosen and lastchosen tables
- when choosing random
  - if the lastchosen table for a user is empty, it gets created. I've tried to keep the old reservedChars logic
  - we then choose a random item from the table and remove it
    - if the item is the same as the last item chosen, we try to choose the item before it. This avoids showing the same character more than once is a bit of an eyesore

### DIFF
--- a/external/script/start.lua
+++ b/external/script/start.lua
@@ -20,7 +20,7 @@ local t_aiRamp = {}
 local t_gameStats = {}
 local t_recordText = {}
 local t_reservedChars = {{}, {}}
-local t_unchosenChars = {{}, {}}
+local t_unchosenRandChars = {{}, {}}
 local timerSelect = 0
 local cursorActive = {}
 local cursorDone = {}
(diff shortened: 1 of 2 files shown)


K4thos:
There are some conflicts after merging #2773, but that PR already added the random selection work in cycles, so the missing feature is already covered. I don’t think adding a new config option is necessary, but let’s discuss it on dscord if you think it’s worth implementing.

potsmugen:
It wasn't really a missing feature. See:
https://discord.com/channels/233345562261323776/233363722934943744/1426768040187400245
So if anything we're less like Mugen now.

PR #2776, opened 2025-10-12
fix: Prevent same random from being chosen twice in a row after refreshing pool

formula1:
After a pool is empty, its recreated and shuffled. While this works fine, sometimes the last item of the previous pool is the same item of the next pool. This causes a character to show up for 2 frames which is the purpose of a newly merged commit.

This double selection is most obvious on select screens with only 2 available items.

### DIFF
--- a/external/script/start.lua
+++ b/external/script/start.lua
@@ -536,11 +536,13 @@ function start.stageShuffleBag(id, pool)
 		end
 		start.shuffleTable(t)
 		start.shuffleBags[id] = t
+
 		-- prevent same stage from being chosen twice in a row
-		if start.previousRandStage[id] == t[#t] then
+		if #t > 1 and start.previousRandStage[id] == t[#t] then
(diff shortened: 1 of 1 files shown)


K4thos:
@formula1, I haven’t verified your observation yet, but let’s wait for @rakieldev to review and possibly propose his version of the fix. Having more than one person working on the same thing isn’t optimal and this new implementation is still under dev (we need to implement sszRandom).

formula1:
@K4thos I figured since the issue was closed and the related pull request was merged this was considered solved.

What's the best place to know who's working on what? Discord?

K4thos:
yes, we're actively discussing randomness implementation on discord right now

formula1:
the-dev-room? https://discord.com/channels/233345562261323776/233363722934943744

K4thos:
yes, that's the dev channel where all features are discussed

PR #2791, opened 2025-10-15
refactor: dummyproof lua random

formula1:
While local, the original math.random is fine.
However, if someone wants to use math.random over net play, it should come from sszRandom. instead of trying to explain to new moders that they should use sszRandom, this just replaces it so they don't have to learn about it.

### DIFF
--- a/external/script/main.lua
+++ b/external/script/main.lua
@@ -4,6 +4,18 @@ main = {}
 --;===========================================================
 --; INITIALIZE DATA
 --;===========================================================
+math.randomseed = function(seed)
+	if type(seed) ~= "number" then
+		error("number expected, got " .. type(seed))
+	end
+	if seed % 1 ~= 0 then
+		error("seed must be an integer")
+	end
+	if seed < 0 or seed >= 2^31 then
+		error("seed out of range, must be between 0 and 2^31 - 1")
+	end
+	sszRandomSeed(seed)
+end
 math.random = function(min, max)
 	-- should return a float between 0 and 1
 	if min == nil then
--- a/external/script/main.lua
+++ b/external/script/main.lua
@@ -4,6 +4,34 @@ main = {}
 --;===========================================================
 --; INITIALIZE DATA
 --;===========================================================
+math.random = function(min, max)
+	-- should return a float between 0 and 1
+	if min == nil then
+		-- int32 maximum is (2^31 - 1)
+		-- using 2^31 should make 1 impossible
+		-- lua's max integer apparently is 2^63 - 1
+		return sszRandom() / 2^31
+	end
+	-- supports math.random(2)
+	if max == nil then
+		max = min
+		min = 1
+	end
+
+	local range = max - min + 1
+	if range <= 0 then
+		error("max must be greater than or equal to min")
+	end
+	if range == 1 then
+		return min
+	end
+	-- sszRandom doesn't support values greater than 2^31
+	-- might as well fail so users don't run into unexpected behaviour
+	if range > 2^31 then
+		error("range must be less than or equal to 2^31")
+	end
+	return min + (sszRandom() % range)
+end
 math.randomseed(os.time())
 
 main.flags = getCommandLineFlags()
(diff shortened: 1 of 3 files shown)


K4thos:
I'm fine with this change since it will also affect external modules, but please make sure to cover all instances that were recently updated to use sszRandom. If we decide to stick with overloading math.random, it should be used consistently everywhere, rather than mixing sszRandom and math.random.

rakieldev:
I’m not too sure about this PR, normally we don’t want Lua running during netplay, and at least I haven’t seen any external module that does run during netplay and specifically uses math.random

formula1:
My understanding was the original shift from math.random to sszRandom was due to the select screen synced over netplay. So there was a need to have synced randoms in lua.

While the only necessary synced random generation may be character, stage and pallet (random background graphics/audio could use local randoms), if a modder wants to include more synced randomizing they would have to learn to use sszRandom. For example, randomizing assist type like in Marvel vs Capcom. Though I'm not sure if Ikemen-Go supports custom configuration like that yet though. Learning sszRandom isn't that difficult but they have to know at some point they have to use it. So it would have to be in an "important" section in Ikemen documents that they should use `sszRandom()` instead of `math.random()`.

Its small but I think it'd be nice if theres one less thing to learn.

formula1:
I'm not sure what to do about setting seeds with sszRandom, I expect they can just be set to 0
https://github.com/ikemen-engine/Ikemen-GO/blob/develop/external/script/main.lua#L1539
https://github.com/ikemen-engine/Ikemen-GO/blob/develop/external/script/main.lua#L2791
https://github.com/ikemen-engine/Ikemen-GO/blob/develop/external/script/main.lua#L2805
https://github.com/ikemen-engine/Ikemen-GO/blob/develop/external/script/main.lua#L3500

formula1:
Something to consider that the Go random is slower than the original random

```lua
-- run before and after overwriting math.random
function runRandTest(randType)
	startTime = os.clock()
	for i = 1, 10000 do
		math.random()
	end
	elapsedTime = os.clock() - startTime
	print("Elapsed time " .. randType .. " Float: " .. elapsedTime)

	startTime = os.clock()
	for i = 1, 10000 do
		math.random(i)
	end
	elapsedTime = os.clock() - startTime
	print("Elapsed time " .. randType .. " Single Int: " .. elapsedTime)


	startTime = os.clock()
	for i = 1, 10000 do
		math.random(i + 1, i * 2)
	end
	elapsedTime = os.clock() - startTime
	print("Elapsed time " .. randType .. " Two Int: " .. elapsedTime)
end
```



```
Elapsed time Original Float: 0.002024785000000001
Elapsed time Original Single Int: 0.0029567789999999927
Elapsed time Original Two Int: 0.0026473220000000075
Elapsed time Go Float: 0.0030214299999999916
Elapsed time Go Single Int: 0.006736154999999994
Elapsed time Go Two Int: 0.01107572200000001
```

rakieldev:
> Though I'm not sure if Ikemen-Go supports custom configuration like that yet though

Only with modules, yeah. 

I get your point, but to me, it feels like you're just wrapping one function around another. If someone already knows how to make modules, using sszRandom directly isn’t really hard. And once the Lua functions are documented after the refactor, it should be even easier 

Well, just my take though

potsmugen:
With rollback netcode the desyncs are very obvious when they happen. Maybe it's a little cleaner to let math.random act as one would expect but just keep in mind sometimes we need the seeded random. On that note I wonder if during matches touching the seed in Lua has the potential to desync char code.

Jesuszilla:
I personally am all for the idea of this as it can prevent a lot of errors from existing Lua users who have prior knowledge. They shouldn't be screwed over just because they don't know about `sszRandom()`.

rakieldev:
Is there any module that runs during netplay though? usually we don't want that

potsmugen:
You shouldn't make any that interfere with gameplay, but other things should be allowed. We had this conversation recently on Discord. If Lua mutates the seed it'd be a big argument for keeping math.random().

K4thos:
Thanks for the feedback, everyone. Let's stick with the current implementation. If we get reports that randomness is still unreliable during netplay, we’ll revisit.

PR #2842, opened 2025-10-27
feat: attackmul and defencemul triggers; fix: Persistent parameter Behavior to Match MUGEN

### LINKED ISSUE #2840: Explod layering regression
### Describe the bug

The drawing order of some explods changed.

I suspect it's the new "sort slice" conditions because it's a very recent issue. This October 20th build I have doesn't have the bug.

### To Reproduce

1. [Download my Pocket Akuma](https://www.mediafire.com/file/p9thv602uwxqzn4/pocshingouki.zip/file)
2. Teleport (DP + 3P/3K)

### Expected behavior

Red aura drawn behind character.

### Screenshots / Video

Image

25/10/2025

### Operating system

None

_No response_

NeatUnsou:
・Corrects the behavior of the persistent SCTRL parameter to accurately replicate MUGEN's specifications, especially during hitpause. persistent counters are inherited across ChangeState during hitpause, just like in MUGEN.

Fixes https://kamekaze.world/neo/

・Implemented MUGEN's quirk where a ChangeState in hitpause disables the SCTRL at the same index in the destination state until the hitpause ends.

(These behaviors are not reproduced when ikemenversion is set)

・Add attackmul trigger and defencemul trigger
These separate the attackmul and defencemul values from the before attack and defence triggers. Accordingly, the attack and defence triggers will no longer include attackmul and defencemul in their calculations. To output the same value as the before triggers, use something like attack * attackmul

・Renamed p2stand.friction and p2crouch.friction to stand.friction and crouch.friction.

・Removed grouping by syncId in draw sort, as it's probably unnecessary (syncLayer is sufficient)
Fixes #2840

### DIFF
--- a/src/anim.go
+++ b/src/anim.go
@@ -1070,10 +1070,6 @@ func (dl DrawList) draw(cameraX, cameraY, cameraScl float32) {
 		if dl[i].priority != dl[j].priority {
 			return dl[i].priority > dl[j].priority
 		}
-		// Then by SyncID to group synchronized sprites together
-		if dl[i].syncId != dl[j].syncId {
-			return dl[i].syncId < dl[j].syncId
-		}
 		// Sort by syncLayer to ensure proper layering within a sync group
 		return dl[i].syncLayer > dl[j].syncLayer
 	})
--- a/src/bytecode.go
+++ b/src/bytecode.go
@@ -2877,7 +2877,8 @@ func (be BytecodeExp) run_ex(c *Char, i *int, oc *Char) {
 	case OC_ex_spriteplayerno:
 		sys.bcStack.PushI(int32(c.spritePN) + 1)
 	case OC_ex_attack:
-		sys.bcStack.PushF(float32(c.gi().attackBase) * c.ocd().attackRatio)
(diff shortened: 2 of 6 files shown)


potsmugen:
Aren't `attack` and `defense` stable version triggers? This would be a breaking change. Even regardless, their name alone implies the final value of attack/defense.

NeatUnsou:
Previously, the attack and defence values specified in Const were immutable, but since ModifyPlayer now allows for effective changes, there has arisen a need to reference the attack value, defence value, attackmul value, and defencemul value individually.
These are distinct concepts, and handling only the final values makes them inconvenient to work with.

That said, it certainly lacks compatibility. Should we limit the previous behavior to cases where ikemenversion < 1, or create new triggers for referencing attack value and defence value?

potsmugen:
Trigger behavior varying by version should really be a last resort thing in my opinion. Too confusing. Instead of changing `attack` and using `attack * attackmul`, `attack` could stay the same and `attack / attackmul` would return base value.

NeatUnsou:
Hmm, that seems better. There are fewer changes.

PR #2851, opened 2025-10-30
feat: add alpha argument to all font fields and Text SCTRL color field (sprite & TTF)

Jesuszilla:
* Adds alpha parameter to all font fields (Lua, lifebar, and Text SCTRL). Tested working with both TTF and sprite fonts.
* Lua `textImgSetColor` modified to take an optional 5th argument that defaults to 255 if not specified for compatibility.

system.base.def contains the information for using the new field (appended to the end of the current font parameters, so alpha comes after height, which is unused for sprite fonts)

Text SCTRL example:
```sh
Text{
	removetime: -1;
	localcoord: FightScreenVar(info.localcoord.x), FightScreenVar(info.localcoord.y);
	layerno: 2;
	linespacing: 25;
	text: "https://ikemen-engine.github.io";
	font: F 0;
	color: 255,255,255,96; # <<< last arg here (optional, defaults to 255 if not specified)
	bank: 0;
	align: 0;
	pos: FightScreenVar(info.localcoord.x)/2, FightScreenVar(info.localcoord.y) - 48.0;
	ID: $txtId;
}
```

Note: Does not support src/dst blending due to TTF limitations and (more important to maintain the same set of features for both), only straight alpha from 0-255 is supported


K4thos:
If alpha is meant to be the font parameter for screenpacks and lifebars, I think we should make it the 7th parameter instead of the 8th. It feels more natural to have the r, g, b, a values grouped together.

Right now, the 7th value is used for the TTF height, but the only place it is documented is system.base.def. I think most people are not aware of what it does considering it's not available for lifebars at the moment and not used in the default screenpack. I think it’s worth changing the order even if there’s a risk that some existing screenpacks might have relied on old one.

Jesuszilla:
> If alpha is meant to be the font parameter for screenpacks and lifebars, I think we should make it the 7th parameter instead of the 8th. It feels more natural to have the r, g, b, a values grouped together.
> 
> Right now, the 7th value is used for the TTF height, but the only place it is documented is system.base.def. I think most people are not aware of what it does considering it's not available for lifebars at the moment and not used in the default screenpack. I think it’s worth changing the order even if there’s a risk that some existing screenpacks might have relied on old one.

Understood. I actually prefer it that way too, reordered.

Issue #1434, opened 2023-10-06
SD Bluestreak Zeus super doesn't react propely after one use.


FrankSP7ps:
### Describe the bug

SD Bluestreak in Ikemen go has apparently one little bug with his super "Zeus".

That little bug being:
After you pressed correctly the strong punch button right when he hits the opponent, he would normally execute his last part as a reward for fill those 3 moments.

If you try to use "Zeus" super again and try to do the same steps...They don't work anymore. Not even if you quit the match, enters another, nope. Until you quit the game and enter or do "shift +F4".
Making impossible for the player or cpu to want to execute right the super.

### To Reproduce

SD Bluestreak from Balthazar old page (Where his non Z2 or DBZ creations are hosted): https://network.mugenguild.com/balthazar/char_bluestreakds.html

Try it with any normal opponent (KFM, Ryu, even with pocket or small characters like Pocket  Athena by Limitedmoon or Shanghai Doll by RoysquadRocks), so the Zeus super executes correctly.

Get 3 bars, and execute the Zeus power by yourself.

Just in case there is a readme where it explains how to do it, and how do the "strong punches" thing.

If you did right, then try it again and you will notice.

### Expected behavior

After you executed perfectly the super, for the second time it would be impossible to do it right.

Try it on Mugen 1.0/1.1 (beta) and you will see that it still works even after one use.

### Screenshots / Video

https://github.com/ikemen-engine/Ikemen-GO/assets/108790388/edf1c3ac-2258-4863-a40a-7150cc35ea55

Im just gonna say that ignore the very crappy index of the buttons as are taken from the HDBZ project (specifically the unnoficial Ikemen version that containts a option to show buttons).
Didn't fixed it as I not use this, besides that Im only using that as a way to show that the Zeus super doesn't work propely after one use as I already said. 

Showing it without that would be difficult, well unless you test it by yourself.

### Engine Version (or source code date)

v0.99.0

### Operating system

Windows

potsmugen:
I have a feeling this is the "persistent + ignorehitpause" bug we've talked about in the Discord channel.

The good news is that it also happens in 0.98.2, so it's not a regression.

FrankSP7ps:
> The good news is that it also happens in 0.98.2, so it's not a regression.

Oh yeah I did discover the bug in that version too, but I did forget to share it a the moment. :P

 > In the meantime you can add "ikemenversion = 1" to the char's def file to fix it.
 
 Did tried it and hey!
[...]

PR #2888, opened 2025-11-17
fix: persistent value when changestate during hitpause

### LINKED ISSUE #2881: A supermove of Neptune by Nep Heart did not work properly in latest nightly
### Describe the bug

The supermove "Victory Slash (D,F,x+y)(D,F,y+z)(D,F,x+z)" of Neptune by Nep Heart did not work properly in latest nightly, which worked perfectly in 2025.08.25 nightly build, also in mugen 1.1. All the test was from a clean default ikemen setting.

https://github.com/user-attachments/assets/99036a82-51df-4834-b21f-9008d7e8ce6a

This is in 2025.08.25 nightly build, mugen 1.1 performs the same. 
Also confirmed this works perfectly in 25.09.10 and even an old 24.03.31 nightly build.

https://github.com/user-attachments/assets/2e37921a-d9b3-4033-9b67-3e7d24a725fc

This is in the latest (2025.11.13) nightly build, she did not perform the final hit at all. When she hit an airborne opponent, she will perfrom the final hit action very quickly but still not hit at all.

### To Reproduce

1. Download the character from https://www.mediafire.com/file/ep0u4weyaqh2wpr/N_Neptune.rar/file (Including two versions, either version is ok)
2. Perfrom 236 x+y and hit a opponent.
3. The supermove performs like in the first video, which was not proper.

### Expected behavior

Perfroming the supermove like in earlier nightly also in mugen 1.1. Thanks very much!!

Latest nightly (2025.11.13)

### Operating system

Windows

_No response_

NeatUnsou:
・Fix #2881 persistent value when changestate during hitpause

1.Carrying Over a Counter to a persistent = 0 SCTRL: 
When a SCTRL with persistent = 0 inherits a persistent counter from a previous state that was still counting down (value was > 0), the new SCTRL will now be treated as if it has already been executed. Its counter is immediately set to a locked state (math.MaxInt32), preventing it from running until the character's persistent counters are next reset. If the inherited counter was already ≤ 0 (ready to execute), the SCTRL remains executable.

2.Carrying Over an Executed persistent = 0 Counter to a persistent > 0 SCTRL:
 Conversely, when an SCTRL with a repeating persistent value (e.g., persistent = 32) inherits a locked counter (math.MaxInt32) from a persistent = 0 SCTRL that had already run, the counter is now reset to the new SCTRL's persistent value. This unlocks the SCTRL.

・Fix Hitdef keepstate
Damage handling for HitOverride
Moved the process of pausing the opponent to system.zss

### DIFF
--- a/data/system.zss
+++ b/data/system.zss
@@ -3,12 +3,20 @@
 #===============================================================================
 [StateDef -4]
 
-# For backward compatibility, x and z-axis get hit acceleration must be handled here
-# TODO: These cases should maybe use the state number constants
-# TODO: These patches could be done by appending sctrl's to the common1.cns states
 if pauseTime = 0 {
+	# For backward compatibility, x and z-axis get hit acceleration must be handled here
+	# TODO: These cases should maybe use the state number constants
+	# TODO: These patches could be done by appending sctrl's to the common1.cns states
 	if time > 0 && (stateNo = 5030 || stateNo = 5035 || stateNo = 5040 || stateNo = 5050 || stateNo = 5071 || stateNo = 5200) {
 		velAdd{x: getHitVar(xAccel); z: getHitVar(zAccel)}
+	}
+	# When keepstate attack, pause at hitshaketime are handled here
+	# TODO: will change the pause process from hitpause to a new pause SCTRL
+	if gethitvar(keepstate) && gethitvar(hitshaketime)>0 && hitpausetime=0 && !HitOverridden {
+		ModifyPlayer{hitpausetime:gethitvar(hitshaketime)}
+		ignorehitpause{
+		GetHitVarSet{hitshaketime:0}
+		}
 	}
 }
 
--- a/src/bytecode.go
+++ b/src/bytecode.go
@@ -590,6 +590,7 @@ const (
 	OC_ex_gethitvar_guardflag
 	OC_ex_gethitvar_stand_friction
 	OC_ex_gethitvar_crouch_friction
+	OC_ex_gethitvar_keepstate
 	OC_ex_ailevelf
 	OC_ex_animelemvar_alphadest
 	OC_ex_animelemvar_angle
@@ -709,7 +710,6 @@ const (
 	OC_ex_prevmovetype
 	OC_ex_prevstatetype
 	OC_ex_reversaldefattr
-	OC_ex_airjumpcount
 	OC_ex_envshakevar_time
 	OC_ex_envshakevar_freq
 	OC_ex_envshakevar_ampl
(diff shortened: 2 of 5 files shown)


potsmugen:
Do Ikemen chars need the `persistent` changes? They were already handling it in a different way before and Mugen's method seems too buggy and unpredictable to try to maintain.
The system.zss block also kind of looks like something the interested people could do in the char without us having to be responsible for it.

NeatUnsou:
IKEMEN chars have never been affected by this specification.
With ChangeState, all persistent values are reset. It's clear.

The behavior where MUGEN chars inherit the previous persistent value is almost a bug, but some chars will not function properly without this, so this process is necessary for compatibility.


The intended purpose of hitshaketime is to temporarily pause the opponent, so there is no reason to give authors discretion over it. (If doesn't want the pause, can set pausetime = 0)
Moreover, to maintain consistency with HitOverride (which uses hitshaketime for pausing), this needs to be handled in the common zss. Otherwise, there will be unintended differences in whether pausing occurs depending on whether HitOverride is used or not.
Removing the common processing would be equivalent to forcing  authors who want consistent behavior to implement hitshaketime pausing themselves, so providing it in advance is the reasonable approach. 
The only concern is that it currently uses hitpause, but I plan to eventually replace it with a safer pause SCTRL.

potsmugen:
It's just too many workarounds. The native `keepstate` implementation seems more complex than ZSS codes that attempt the same thing manually, ironically. But this is just one opinion.

NeatUnsou:
original common.cns also does not hardcode the pause processing, it is written in the cns instead. (though it uses hitshakeover instead of hitshaketime)
so I think this approach is fine.

K4thos:
@NeatUnsou, for readability sake please wrap this code with the engine version checks even if they are redundant.

NeatUnsou:
Don't worry, version checking already works
```
	if c.stWgi().ikemenver[0] != 0 || c.stWgi().ikemenver[1] != 0 {
		c.ss.sb.ctrlsps = make([]int32, len(c.ss.sb.ctrlsps))
	} else {
		// Reset persistent counters for this state (MUGEN chars)
		if c.hitPause() && ctrlsps_backup != nil {
			// If changing state during hitpause, restore (carry over) persistent from the before state
			c.ss.sb.ctrlsps = make([]int32, len(c.ss.sb.ctrlsps))
			copy(c.ss.sb.ctrlsps, ctrlsps_backup)
			// Apply special persistent counter behavior for state changes during hitpause
			c.persistentChangeStateHitpauseCorrection(&c.ss.sb)

			// Get the index of the currently executing SCTRL block
			c.hitStateChangeIdx = c.currentSctrlIndex
		} else {
			// If not in hitpause, reset persistent
			c.ss.sb.ctrlsps = make([]int32, len(c.ss.sb.ctrlsps))
			c.hitStateChangeIdx = -1
		}
	}
```

NeatUnsou:
I think #1434 has also been fixed.
I've tested it and the reported issue did not occur.

PR #2999, opened 2025-12-24
feat: motif and lifebar elements projection support

rakieldev:
Feat:
- All motif and lifebar elements (sprites and TextSprites) now support the following parameters:
``xangle``, ``yangle``, ``projection``, and ``focallength``.
- Transformations can be applied to cells, and these transformations are inherited by portraits, cell bgs, and the cursor.
- It is now possible to edit entire columns or rows of cells.
An asterisk can be used as a wildcard to apply changes broadly:

```ini
;<col-*> for columns
;<*-row> for rows
```

```ini
cell.0-*.xshear = 0.5
```
Applies xshear to all cells in the first column.

```ini
cell.*-0.offset = -30, 0
```
Changes the offset of all cells in the first row.

Fix:
- Fixed an issue where animations copied for the palmenu were not applying some transformations


K4thos:
@rakieldev, in motif.go
Projection string `ini:"projection"`
should be changed to
Projection string `ini:"projection" default:"orthographic"`
(so that it works with correct default even if the value is not set in defaultMotif.ini)

rakieldev:
> @rakieldev, in motif.go Projection string `ini:"projection"` should be changed to Projection string `ini:"projection" default:"orthographic"` (so that it works with correct default even if the value is not set in defaultMotif.ini)

@K4thos Done, added the missing defaults to motif.go.

PR #3100, opened 2026-01-09
fix: Unassigned menu input coming up with A button

### LINKED ISSUE #3091: A button opens menu even when disabled ("Not Used") in gamepad config
### Describe the bug

After @Jesuszilla's SDL PR #2908, I can reproduce an issue where the A button opens the menu even when the menu button is explicitly disabled in the gamepad input config (set to “Not Used” via Spacebar while binding).

This is not related to my recent refactoring - the same behavior occurs on an engine state from before my changes.

### To Reproduce

1. On Windows, connect a controller.
2. Start the match with kfm.
3. Open Gamepad input config.
4. Highlight menu button and press Spacebar to set the binding to “Not Used” (disabled).
5. Return to the match.
6. Observe that A button opens the menu despite Menu button being disabled.

### Expected behavior

If a button is set to Not Used, pressing it should do nothing for that action.

### Screenshots / Video

_No response_

nightly post SDL changes

### Operating system

Windows

_No response_

Jesuszilla:
Fixes #3091

### DIFF
--- a/src/input.go
+++ b/src/input.go
@@ -3720,7 +3720,7 @@ func (cl *CommandList) IsControllerButtonPressed(token string, controllerIdx int
 	}
 	// Non-analog controller tokens (A/B/X/Y/etc.): simple digital button check.
 	idx, ok := StringToButtonLUT[token]
-	if !ok {
+	if !ok || idx == 25 { // if not successful or "Not used" is ever returned
 		return false
 	}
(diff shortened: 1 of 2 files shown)


K4thos:
This "Not used" string can be translated. So, either use `sys.motif.OptionInfo.Menu.Valuename["nokey"]` to retrieve the actual string, or fall back to the default case so that any unrecognized string is ignored.

Jesuszilla:
Making a new PR with more fixes

PR #3159, opened 2026-01-20
fix: set correct gamemode in arcade, teamarcade, versus and teamversus modesprovide correct gamemode in arcade, teamarcade, versus and teamversus…

updeshxp:
Ikemen Go currently does not provide correct value in gamemode() for team modes, Below have been corrected:
teamarcade mode was also being reported as arcade mode.
teamversus mode was also being reported as versus mode.

In this PR, correct mode is provided to allow external and internal lua modules to detect mode correctly.

### DIFF
--- a/external/script/main.lua
+++ b/external/script/main.lua
@@ -1758,6 +1758,7 @@ main.t_itemname = {
 			main.teamMenu[2].single = true
 			textImgSetText(motif.select_info.title.TextSpriteData, motif.select_info.title.text.arcade)
 			main.teamarcade = false
+			setGameMode('arcade')
 		else --teamarcade
 			main.teamMenu[1].ratio = true
 			main.teamMenu[1].simul = true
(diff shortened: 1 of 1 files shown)


K4thos:
The way it works now is intentional. There are other ways for Lua scripts to detect which mode was selected. I need to think through all the implications of a change like this since there was a reason (which I don't remember right now) why they are shared.

updeshxp:
In start.lua,
There is a condition currently, which also depends on gamemode as teamversus, could you please check on that as well:
-> line 407 - 
`if start.challenger > 0 and gamemode('teamversus')`

rakieldev:
Well, essentially teamarcade is just normal arcade but with the teammenu enabled same for versus. We also have ``main.teamarcade = true``. which is set inside the block that enables the teammenu at least for arcade, which kind of loses its meaning with this addition I guess. 

The thing is, these modes are still ``arcade`` and ``versus``, just with different teammodes enabled, and arcade and teamarcade shares the same hiscore unless they’re deliberately separated. That might be one of the reasons, they're shared?

K4thos:
yeah, Rakiel is right, that was the reasoning. I will change that  gamemode('teamversus') to gamemode('versus') in code

PR #3189, opened 2026-01-26
feat: add getStoryboardScene() and scene jump support to storyboards

dionednd:
Add getStoryboardScene() lua function to expose the currently active storyboard scene.
This allows lua scripts to make decisions based on the current scene, like handling vn choice results and dynamic scene jumps.

The change is fully compatible with existing storyboards.


dionednd:
This PR also adds a new SceneDef parameter: disablecancel (boolean).

PR #3279, opened 2026-02-14
feat: MatchRestart resetmatch preservevars pXpalette

NeatUnsou:
Added new parameters for MatchRestart

・resetmatch
If set to 1, the match will restart from Round 1 instead of the current round. 
・preservevars
This parameter specifies whether to preserve vars for a particular character after the restart. 
・pXpalette
Specifies the palette number to be used for player X (need to reload)

・Removed the +1 increase in supermovetime and pausemovetime for Explod. This only applied to projectile.


K4thos:
@NeatUnsou, there is a merge conflict

PR #3286, opened 2026-02-16
feat: AssertSpecial nodestroyself; PlayerPush affectteam

NeatUnsou:
Added a new flag to AssertSpecial and PlayerPush

AssertSpecial 
- nodestroyself
Disables destroyself while this flag is active.

PlayerPush
- AffectTeam
F ; Allows only allies to be pushed out (enemies will pass through)
B ; Pushes both allies and enemies
E ; Pushes only enemies (default)

### DIFF
--- a/src/bytecode.go
+++ b/src/bytecode.go
@@ -9754,6 +9754,7 @@ type playerPush StateControllerBase
 const (
 	playerPush_value byte = iota
 	playerPush_priority
+	playerPush_affectteam
 	playerPush_redirectid
 )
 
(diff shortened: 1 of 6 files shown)


potsmugen:
Maybe `teampush` could be a PlayerPush parameter. Even with the Hitdef `affectteam` syntax maybe. We already expanded PlayerPush with things like priority.

NeatUnsou:
> Maybe `teampush` could be a PlayerPush parameter. Even with the Hitdef `affectteam` syntax maybe. We already expanded PlayerPush with things like priority.

I changed it to AffectTeam in PlayerPush instead of Teampush. Thank you for the idea.

PR #3377, opened 2026-03-07
fix: Removed OpenGL 2.1 setting and hack docs: Added comment about Mugen RemapPal inconsistency

jenngra505:
This fixes a crash when selecting a menu setting that was meant to be removed during a refactor.
I also added a comment in the sff palette search section of the bytecode to point out an unfixable inconsistency with Mugen based on what was discovered in https://github.com/ikemen-engine/Ikemen-GO/issues/168


K4thos:
6 commits when the change is supposed to remove 1 line. Don't think you should touch shaders code unless you completly understand what that removed line actually does (I don't so not touching it either). I'll remove defaultMotif value in next push.

PR #3404, opened 2026-03-14
fix: screenpack fonts, lifebar refilling; refactor lifebar drawing order

potsmugen:
Fixes:
- Fixed lifebar front layer having a refill lag when life is increased rather than depleted
- Fixed red life drawing the whole rectangle instead of just the difference between red life and life
- Fixed TextSprite being scaled incorrectly in widescreen (regression)

Refactor:
- Inverted lifebar drawing order so that player 1 ends up on top, like most of our drawing loops

### DIFF
--- a/src/font.go
+++ b/src/font.go
@@ -1267,13 +1267,15 @@ func (ts *TextSprite) Draw(ln int16) {
 
 		// Adjust draw scale to the font's original localcoord
 		// We only multiply by localScale here because SetScale divides by it
-		scaleRatio := 320.0 / float32(ts.fnt.localcoord[0]) * ts.localScale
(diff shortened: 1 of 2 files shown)


K4thos:
wouldn't this cause the issue I've mentioned here? https://discord.com/channels/233345562261323776/1466719790805684256/1482284487630127104
May be better to make reverse rendering order a toggle

potsmugen:
The reverse rendering itself should be safe. It just does what we already do, but puts player 1 in front of player 2 (and so on).
The change you're replying to on Discord was just a test that's not in this PR. It would draw all of a player's sprites before moving on to the lifebar of the next player.

PR #3431, opened 2026-03-24
fixes: Stage edges break bug

NeatUnsou:
~~・ Added -debugapi option~~
This allows IKEMEN GO to interact with other software via an API. This functionality is necessary to create a debugging tool for IKEMEN GO that offers the same level of convenience as the MUGEN debugging tool, "窓無限の十徳ナイフ (Swiss Army Knife for winmugen)".
By entering the option "-debugapi [port number]" and enable debug mode, IKEMEN GO will function as a TCP server and can send JSON 60 times per second. The default port number is 8080.

This test tool will allow to view character information.
ikego-tool_view
[Ike-Spect.zip](https://github.com/user-attachments/files/26225177/Ike-Spect.zip)

・ -speedtest option speed can be specified from the command line 
・ Add -framerate option

・ Fixed an issue where the edges of the stage would break when game speed and framerate were set to extreme levels.


K4thos:
Under discussion: https://discord.com/channels/233345562261323776/233363722934943744/1486112122030264477
Summary: ideas for a Lua equivalent, as well as a different approach to exposing triggers without the need to maintain the API. PotS also prepared a different solution for VarSet, we need to decide which fix is better.

potsmugen:
My fix is truer to Mugen in that it allows anything between `var(x)` and `=`, rather than just `:`.
I was also responsible for the regression a few days ago. Normally we use Discord for these things so our work doesn't overlap. It's not like a made the problematic commit a long time ago.

NeatUnsou:
I'll revert the change to ':=' in VarSet. @potsmugen's method is more accurate to MUGEN.

NeatUnsou:
I considered using Lua, but I felt it was too resource intensive compared to the API, which uses a separate thread to mitigate the impact on operation.
The API simply reads parameters and creates a dump within Go, whereas with Lua, Lua and Go must communicate for each required parameter before creating dump. The speed difference is obvious.
and, directly writing to parameters would require creating many new functions in script.go. There is no suitable functionality in existing IKEMEN GO to do this.

I believe the API is the most practical approach for replicating the debugging features of SAK and ShugenDo.

potsmugen:
I also fixed the `var(=)` bug, in case that was ambiguous. My fix is very similar to yours anyway, but that's just going to create merge conflicts.

Regarding debug, I think the new "teamside" commit is already showing the problem with this kind of approach.

NeatUnsou:
Okay, I'll revert the changes to var(=) as well

Regarding parameter referencing, I don't see much of a problem with the current implementation, as it's similar to script.go.
Most triggers simply reference variables, and triggers involving complex processing are already function based, like p2BodyDistX().
However, there are certainly gray area triggers that aren't complex enough to warrant a function but require some calculations.
If someone suggested that the processing of these gray area triggers should also be function based, I would agree, as that's where mistakes are likely to occur.

potsmugen:
They are a gray area indeed because for instance having a helper for TeamSide would still result in confusion because you want the +1 index offset sometimes but not all the time. In this particular case explicit handling is probably clearer.

But I mostly meant that having Ikemen take care of the whole translation layer is bound to be fragile like that.

NeatUnsou:
The only way to avoid having a translation layer on the IKEMEN GO side would be to use a memory editor approach, but this is not suitable for IKEMEN GO because it would be difficult to keep up with engine updates.

potsmugen:
It's kind of the same situation as characters breaking all the time with Ikemen nightly changes. Content shouldn't have to adapt to our beta versions, just stable versions.

NeatUnsou:
There's some truth to the idea that "only need to support stable versions," but the more versions support, and the easier it is to support them, the better things become.
If I were to create a memory editor style debugging tool, it would take many times more time and effort than it took me to create this PR and test tool. Moreover, even if it were completed, it would quickly become outdated.
This explains why an debugging tool for IKEMEN hasn't existed until now.
I've seen creators refuse to migrate from winmugen because SAK didn't support MUGEN 1.1/IKEMEN. it took over 8 years for SAK to support MUGEN 1.1

K4thos:
So the main issue is that this PR adds support for an external app, but it also means we’d have to maintain code that the engine itself doesn’t really use. And it sounds like the general agreement is that supporting stable releases is enough.

Maybe there’s a middle ground: we merge it shortly before 1.0, then revert it once nightly releases are back. Then next time a stable release is coming up, we can add the updated code again (or not, if Neat doesn’t provide an updated version). We could keep this code in a separate branch.

How does this sound @NeatUnsou @potsmugen @rakieldev?

NeatUnsou:
OK, I will temporarily withdraw the implementation of debugapi.
Other fixes will remain, so please go ahead and merge those.

However, I have not given up on having the necessary features for debugging tools built in by default. Next time, I’d like to aim for more maintainable code while preserving the functionality I proposed this time.

PR #3454, opened 2026-03-29
Sblib setup

PeterDunlop:
Set up SBLib with Step and Config. Right now the mutation on the in game, game state doesn't work. We will talk about it more on monday or tuesday. So SBLib takes the gamestate, adds the activation functions to the state, but it doesn't update the game state in the actual game.

To test it check external/config/mods

First see if config-example.lua makes sense. Then run next.

Run a test server (i recommend dummyhttp) with verbose: 
dummyhttp -vv -p 3000 

Then run the game
./Ikemen_GO_Linux   

Check if the post recost for the config looks right in the server. It should send the response with the payload.
To test the step function you should just test that the post request has a body in the dummyhttp. Right now it will crash and its disabled since the server returns nothing, but it should still provide the body for the http if sent.


K4thos:
184 files changed? We can't merge a pr in a state like this. As for the feature, I don't really understand what is it about and why it's needed

PeterDunlop:
My apologies I've accidentally created a PR for the main repository, my intentions were to create a PR for a forked repository. Please disregard this PR.

PR #3483, opened 2026-04-04
fix: Negative sprite numbers not accurate to Mugen, compiler error

### LINKED ISSUE #3232: When some characters use animation group numbers and indices that exceed 65535 (the maximum limit of uint16), the animations with indices beyond this threshold won’t display in Ikemen GO
### Describe the bug

When some characters use animation group numbers and indices that exceed 65535 (the maximum limit of uint16), the animations with indices beyond this threshold won’t display in Ikemen GO — yet they show up properly in MUGEN 1.1

### To Reproduce

https://www.mediafire.com/file/rzayotud6f0qq7t/Vampire.7z/file
ikemen go
Image
mugen1.1
Image

This character appears to exploit the uint16 overflow bug — as evident here, both the group numbers and indices are showing up as negative values.
Image

### Expected behavior

_No response_

### Screenshots / Video

_No response_

the lastly nightly

### Operating system

None

_No response_

jenngra505:
Fixes:
* Added a workaround that allows negative sprite numbers to display as they would on Mugen, useful for unusually coded compatibility animations.
* Fixes https://github.com/ikemen-engine/Ikemen-GO/issues/3232
* Corrected the makefile pointing to an outdated file name.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,7 @@ srcFiles=src/resources/defaultConfig.ini \
 	src/compiler_functions.go \
 	src/config.go \
 	src/dllsearch_windows.go \
+	src/fightscreen.go \
 	src/font.go \
 	src/font_gl33.go \
 	src/font_gles32.go \
@@ -24,7 +25,6 @@ srcFiles=src/resources/defaultConfig.ini \
 	src/iniutils.go \
 	src/input.go \
 	src/input_sdl.go \
-	src/lifebar.go \
 	src/main.go \
 	src/motif.go \
 	src/music.go \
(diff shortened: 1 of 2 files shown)


K4thos:
wontfix labelled issues don't need fixing

NeatUnsou:
The current implementation of IKEMEN GO already forces int32 values ​​that cannot be handled by uint16 (values ​​exceeding 65535) to be cast to uint16.
I don't think extending this range to negative values ​​would impair functionality. This is more of a change to align the compiler's quirks with MUGEN than an emulation of a bug.
For compatibility reasons, this change should be merged.

PR #3498, opened 2026-04-12
fix: avoid invalid player lookup in IkSys_ReceivedDamage redirect path

### LINKED ISSUE #3497: "found no player with id: " debug spam
### Describe the bug

this bug, caused by helpers or lifeadd with redirectid hitting an enemy in some circumstances, was recently found and solved on discord, but it's still not part of the release, so i'm leaving the information here about it as a reminder.
the problem happens in functions.zss on line 188, it needs a fallback to avoid showing those errors. the new line should read:
`	if receivedDamage > 0 && getHitVar(playerNo) > 0 && playeridexist(gethitvar(playerid)) && playerId(getHitVar(playerId)),teamSide != teamSide
`
to prevent debug messages.

### To Reproduce

1. hit an enemy with a helper that disappears very quickly. this will sometimes cause the debug warning.

### Expected behavior

_No response_

### Screenshots / Video

_No response_

2026.04.11

### Operating system

Windows

_No response_

saschabuehrle:
Fixes #3497

When redirected hit data references a player id that no longer exists (for example short-lived helpers), `IkSys_ReceivedDamage()` evaluated `playerId(...)` directly and spammed debug logs.

This adds a `playerIdExist(getHitVar(playerId))` guard before dereferencing the redirected player, preventing invalid lookup noise while keeping the existing team-side filter logic unchanged.

Greetings, saschabuehrle

### DIFF
--- a/data/functions.zss
+++ b/data/functions.zss
@@ -185,7 +185,7 @@ if map(_iksys_comboCountFlag) = gameTime {
 [Function IkSys_ReceivedDamage() ret]
 let ret = 0;
 if receivedDamage != map(_iksys_receivedDamageCurr) {
-	if receivedDamage > 0 && getHitVar(playerNo) != 0 && playerId(getHitVar(playerId)),teamSide != teamSide {
+	if receivedDamage > 0 && getHitVar(playerNo) != 0 && playerIdExist(getHitVar(playerId)) && playerId(getHitVar(playerId)),teamSide != teamSide {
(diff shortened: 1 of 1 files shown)


K4thos:
this is already fixed in PotS' branch that will be merged soon: https://github.com/ikemen-engine/Ikemen-GO/issues/3497#issuecomment-4231037193
Merging it now would likely result in merge conflict with that pr.

potsmugen:
I fixed it more or less like this first but then realized this case would be reason enough to add a `getHitVar(teamside)` trigger.

PR #3502, opened 2026-04-13
fix: Make PalFX.hue shift hues correctly and fix PalFX on 3D models

Alceris:
The previous formula did very poorly preserving the saturation and value of the colors, and if I recall correctly it wasn't even very accurate on the hue, either. This gives you results exactly like what you'd get in an actual image editor like GIMP, barring floating point imprecision or possibly transparency blending.


Alceris:
image
The reason there's such a huge difference in the background is because PalFX was entirely broken on 3D models before.

Jesuszilla:
Confirmed working performant on Atari VCS (dual core processor, 4GB shared RAM) and Android 16 (Pixel 7, 8GB shared RAM)

Alceris:
Turns out this apparently also fixes the alpha channel on 3D stages being blown the fuck out, which does mean people will have to crank up the opacity on transparent stuff in order to get the results they want, but like, 6.25% opacity should be barely visible, not make for a brightly glowing neon sign.

PR #3509, opened 2026-04-15
fix: Fix AllPalFX and BGPalFX being a frame late.

Alceris:
BGPalFX updates to what it should be while the game is paused, so it's a little tricky to spot that it's a frame late. I only noticed it by accident, myself.

### DIFF
--- a/src/rollback.go
+++ b/src/rollback.go
@@ -220,9 +220,7 @@ func (rs *RollbackSystem) simulateFrame(s *System) bool {
 		return false
 	}
 
-	s.bgPalFX.step()
 	s.stage.action()
-
 	// If frame is ready to tick and not paused
 	//sys.stage.action()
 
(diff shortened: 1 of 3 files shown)


potsmugen:
I think this needs more testing. This doesn't just reorder the functions but also changes their dependencies. `stage.action()` was running all the time but now it's gated by `sys.tickFrame()`. Might be relevant for video stepping for instance.
The reordering itself will also make stage triggers 1 frame late, which is fairly significant for attached characters.

Alceris:
> I think this needs more testing. This doesn't just reorder the functions but also changes their dependencies. `stage.action()` was running all the time but now it's gated by `sys.tickFrame()`. Might be relevant for video stepping for instance. The reordering itself will also make stage triggers 1 frame late, which is fairly significant for attached characters.

Fixed.

PR #3510, opened 2026-04-16
fix: ProjVar TeamSide parity with TeamSide

Alceris:
ProjVar TeamSide was returning 0 and 1, instead of 1 and 2 like the other TeamSide triggers.


K4thos:
same fix needed in script.go

Alceris:
Man, I always forget about script.go, fixed.

PR #3511, opened 2026-04-16
refactor: Rename MissOnReversalDef to IgnoreReversalDef

Alceris:
"MissOnReversalDef" was misleading because it did the exact opposite of miss; it _hit through_ ReversalDef.

I'm leaving it as an int32 instead of a bool because I'm planning on having multiple modes for it depending on the value used.

### DIFF
--- a/src/bytecode.go
+++ b/src/bytecode.go
@@ -7305,7 +7305,7 @@ const (
 	hitDef_stand_friction
 	hitDef_crouch_friction
 	hitDef_keepstate
-	hitDef_missonreversaldef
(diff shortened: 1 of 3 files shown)


K4thos:
looks like missonreversaldef was missing in script.go. Can you please add it as part of this pr?

Alceris:
I'm not actually sure how to do that.

Alceris:
In fact I don't think that's actually applicable, as there's no trigger for it to begin with.

K4thos:
you're right, I thought it's also in gethitvar

PR #3551, opened 2026-04-25
feat: custom shader

NeatUnsou:
A feature that allows change the shader applied to a sprite to a different one.
Furthermore, the texture obtained from the rendering result can also be handled by the shader.

* Shader Loading
Specify from the def file.
The left side is the key name to call from cns/zss, and the right side is the file path.
Currently, the loaded custom shaders are available for use by the entire character, so need to choose a name that won't conflict with another character's.
For Vulkan, an spv file with the same name and the extension .spv added is required.

```
	[Shaders]
	kfm_zss_distortion = distortion.frag
	kfm_zss_wobble = wobble.frag
	kfm_zss_disintegrate = disintegrate.frag
```

Shaders are currently available via ShaderSet SCTRL, Explod, and ModifyExplod.
[kfm_zss_s.zip](https://github.com/user-attachments/files/27082035/kfm_zss_s.zip)

[2026-04-25 19-14-48.webm](https://github.com/user-attachments/assets/7d515e9c-5e14-4487-b3de-53770c22012a)

```
	shaderSet{
		shader :"kfm_zss_distortion";
		shaderparam.p0 :2.0;
	}

	explod{
		anim: 196;
		postype : left;
		pos: 0, 0;
		removetime: 51;
		RemoveOnGethit :1;
		scale :(ScreenWidth)/120,(ScreenHeight)/140;
		Shadow :0;
		ownpal :1;
		under :1;
		shader :"kfm_zss_wobble";
	}

	ModifyExplod{
		ID :192;
		shader :"kfm_zss_disintegrate";
		shaderparam.p0 :gametime/60.0;
	}
```
Because this process replaces existing shaders, need to retain the existing shader functionality when processing sprites.
In the sample kfm_zss_s, the existing shader processing is put into a function, and the sprites are obtained from that function.

In addition to the parameters native to IKEMEN GO, the following parameters are available within the shader.

float iTime:
Indicates the number of seconds since the game started. The value is equivalent to GameTime/60.0.

vec2 iResolution:
Equivalent to the current game resolution.

float aspectRatio:
Normally 1.0. If the aspect ratio is changed via the pause menu, etc., the value changes according to that ratio.

sampler2D bgl_RenderedTexture:
Stores the texture obtained from the back buffer at the time of the call. If "bgl_RenderedTexture" is written in the shader, the process of obtaining the back buffer will be performed (even if commented out).

float p0～p15:
Parameters that can be inserted from cns/zss. They can be specified with shaderparam.pX = value.
In Vulkan, they are defined in push_constant.


gerarreal:
Amazing feature. I hope this gets merged.

potsmugen:
This is neat (no pun intended). I haven't looked at the KFM test case yet but what determines if for instance the shader also affects the stage?
Also we'll probably need common shaders now (like common states etc).

NeatUnsou:
This doesn't actually affect the stage; it simply processes the acquired back buffer with a shader and displays it.
Ikemen_GO010
Turning off the shader results in this display.

A common shader should be implementable eventually. It will probably work similarly to the current fightFx system

potsmugen:
So in the case of explods their own animation won't be rendered? I thought it would be like "draw this explod using this shader". Why use shaders on explods instead of shaderSet in that case?

NeatUnsou:
It all depends on the shader. free to use the original sprite or discard it.
The original animation will play as intended if you write code that functions similarly to the original shader.

Alceris:
I'd like to request making it possible to apply custom shaders to stage elements, as well.

Kasasagi77:
> I'd like to request making it possible to apply custom shaders to stage elements, as well.

That would definitely be great for implementing stage's raster effects, such as e.g. the hot air warping in the Fire at the Wadamoya stage from LB2.

leonkasovan:
> I'd like to request making it possible to apply custom shaders to stage elements, as well.

it's like bring a shadertoy to ikemen stages. 
https://www.shadertoy.com/view/Ms2SD1
{CEA25912-3EC6-4AF1-87C0-C1E89C9A4625}

gerarreal:
I'd like to request if it's possible to have shaders play on the character select screen when a character's sprites are previewed.

Kasasagi77:
I have made an official feature request: #3557. Please upvote :smiley:

PR #3568, opened 2026-05-01
feat: Projectile shader parameter

NeatUnsou:
Added a shader parameter to Projectile.
Added a time parameter to shaderSet.  Set to -1 to display indefinitely. Defaults to 1.


potsmugen:
Explod and projectile will probably need time as well at some point.
And maybe the shader timer could be accessible by the shader code? Similar to iTime.

PR #3578, opened 2026-05-03
feat: Load glTF extras from "meshes" array

SuperFromND:
Recent versions of Blender (e.g. 5.1.1, the latest stable, does this) will export glTF extras placed onto objects in the "meshes" field, rather than directly in the "nodes" field as previous versions (4.2, likely later) would do.

This could prove to be a massive headache for those making 3D stages with more recent versions of Blender (as it would encourage either sticking to older versions or hex/script-editing glTFs afterwards), so I've added a bit of code to copy the Extras over if they're present.

Marked as a feature since this is more futureproofing than anything.

### DIFF
--- a/src/model.go
+++ b/src/model.go
@@ -263,6 +263,7 @@ type Mesh struct {
 	name               string
 	morphTargetWeights GLTFAnimatableProperty // []float32
 	primitives         []*Primitive
+	extras             interface{}
 }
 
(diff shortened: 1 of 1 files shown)


SuperFromND:
Never mind, I was wrong. You can just use the Object Properties tab, not Data Properties, and get exported properties in the same place. My bad!

PR #3601, opened 2026-05-10
fix: Rectangles' diagonal rendering artifacts (rasterization edge case).

### LINKED ISSUE #3583: Simple scaling/offset causes straight horizontal/vertical lines to become jagged/diagonal
### Describe the bug

When a simple rectangular scaling or an offset is applied to a sprite, then purely horizontal/vertical lines should remain perfectly horizontal/vertical. Such lines instead sometimes become sligtly jagged/diagonal, as if a slight rotation or shear or some parallax/projection is applied. This should not happen - it is a rendering artifact.

Additionally, using `StageFit` introduces new/different artifacts to the mix.

These problems are the most visible on bitmap fonts, but they affect all of the rendred sprites.

These problems do not exist in Mugen 1.1 - without `StageFit` all of the lines are perfectly crisp and straight, with `StageFit` there seems to be a slight blurring/filtering instead.

### To Reproduce

Take a fresh Ikemen GO install and set the config to:
```
GameWidth               = 1920
GameHeight              = 1080
Fullscreen              = 1
WindowScaleMode         = 0
```
I set this resolution, because it is the native resolution of my monitor - if the native resolution of your own monitor is different, you may need to take screenshots and zoom in to actually see the problem, because there will be another scaling layer in your setup that may obscure the rendering artifacts.

If checking this problem by taking screenshots, make sure that the size of the screenshot is exactly equal to the game resolution (i.e. no window scaling was applied etc).

Using `WindowScaleMode = 1` does no remove the artifacts - it just blurs them out.

---

Test scenario 1 (default screenpack):
1. Run Ikemen GO.
2. Take a screenshot of the main menu.
3. Zoom in and look at the lower left corner of the screenshot with a jagged "I.K.E.M.E.N GO" text.
4. Open the `system.def` file.
5. In the `system.def` file set:
```
footer.title.scale = 3.001,3.001
footer.info.scale = 3.001,3.001
footer.version.scale = 3.001,3.001
```
6. See how the above hack fixes the problem.

---

Test scenario 2 (test stages):

1. Run the provided test stages (one for 4:3, one for 16:9): [scaling.zip](https://github.com/user-attachments/files/27369258/scaling.zip).
2. Watch the bitmap text looking for the artifacts.
    a. Also check the screenshots below.
3. The presense/absense of the artifacts sometimes depends on the camera's position.
4. The 4:3 variant of the test stage has different artifacts with and without `StageFit`.
    a. Turning on `StageFit` seems to introduce its own, unique artifacts. 

### Expected behavior

Matching Mugen 1.1 behavior without `StageFit` - perfect, crisp and straight rendering.

### Screenshots / Video

Default screenpack.

<table>
  <tr><td>Artifacts/Problem</td><td>Expected/Correct</td></tr>
  <tr>
    <td>Image</td>
    <td>Image</td>
  </tr>
</table>

---

16:9 Test Stage Artifacts.

Image

---

4:3 Test Stage Artifacts (`StageFit` **turned off**).

Image

---

4:3 Test Stage Artifacts (`StageFit` **turned on**):

Note the new artifacts introduced by `StageFit` - the grid pattern is broken on the screen diagonal between the upper right and the lower left triangle - the artifacts on this diagonal are also applied to the text.

Image

2026-05-02

### Operating system

Linux

This is just a wild shot in the dark, wihtoul looking at the code, but maybe the rotation or shear or other transformations are applied to all sprites, regardless, and an accumulation of floating point errors causes the sprites to be just slightly rotated/sheared/distorted when they shouldn't.

The unique `StageFit` related artifacts, on the other hand, probably have a totally different cause/origin.

Kasasagi77:
This fix effectively side-steps a low-level rectangle rasterization edge case, that caused visible and frequent artifacts on the diagonal (fix: #3583).

## FAQ

1. **Did you just add a random number in a random place in the code?**

    Short answer: No :stuck_out_tongue_closed_eyes:

    Long answer: @SuperFromND helped me narrow it down to a low-level rasterization edge case, since it turned out that the vertex positions did form a perfect/undistorted rectangle. The usual solutions to such edge cases involve some kind of texture sampling tricks.

    I tried some more complicated and orthodox solutions first but they did not properly fit the current rendering pipeline and I did not want to make deep changes to the pipeline, nor do I think they would be optimal. Ultimately, I reduced the solution to this minimal "one simple trick" that proved effective in wiping out the artifacts in all of my test cases.

---

2. **Won't this cause distortion in the rendered sprites' texture/graphics?**

    Short answer: Effectively no, it should only affect edge cases.

    Long answer: Assuming the worst case scenario of the maximum size of a sprite: 2^15-1, the distortion (in texels) would only be at the level of: (2^15-1) * 0.000002 = 0.0655, so still way below even one actual texel.

---

3. **Won't this cause gap/margin artifacts, like XYTruncate (#2564)?**

    No. The vertex positions are unchanged and nothing is sampled outside of the original texture/sprite area.

    This PR will however also **not** fix any of the previously occuring artifacts of this type.

### DIFF
--- a/src/render.go
+++ b/src/render.go
@@ -227,11 +227,17 @@ func (rp *RenderParams) IsValid() bool {
 func drawQuads(modelview mgl.Mat4, x1, y1, x2, y2, x3, y3, x4, y4 float32) {
 	gfx.SetUniformMatrix("modelview", modelview[:])
 	gfx.SetUniformF("x1x2x4x3", x1, x2, x4, x3) // this uniform is optional
+
+	// This effectively side-steps a low-level rectangle rasterization edge case,
+	// that caused visible and frequent artifacts on the diagonal.
+	// See: https://github.com/ikemen-engine/Ikemen-GO/issues/3583
+	uvZero := float32(0.000002)
+
 	gfx.SetVertexData(
 		x2, y2, 1, 1,
-		x3, y3, 1, 0,
-		x1, y1, 0, 1,
-		x4, y4, 0, 0,
+		x3, y3, 1, uvZero,
+		x1, y1, uvZero, 1,
+		x4, y4, uvZero, uvZero,
 	)
 
 	gfx.RenderQuad()


potsmugen:
> uvZero := float32(0.000002)

I get a bad feeling about lying to the renderer like this. Like the bug is being swept under the rug and may appear somewhere else. It's fixing a symptom.

Kasasagi77:
> > uvZero := float32(0.000002)
> 
> I get a bad feeling about lying to the renderer like this.

I think there is maybe some misconception because of the way I named this variable.

This just intruduces a minimal bias/distortion so that the edge cases of the rasterizer itself (which is, in fact, implemented by the GPU vendor and/or the OS) are not triggered in a way that causes artifacts.

> Like the bug is being swept under the rug and may appear somewhere else. It's fixing a symptom.

A symptom of what exactly? Where do you propose the bug is?

Don't you feel that sampling the texture from the literal edge (0.0) to the literal other edge (1.0), both edge inclusive, sounds like an off-by-one bug intead? Like iterating from 0 to 10 (both inclusive) and expecting the loop to repeat 10 times? Anyway - this is only an analogy.

potsmugen:
The fact `0.000001` didn't work makes it look like "adjust magic number until it works". But you've looked into this more than I have, so my opinion is just based on general principles/etiquette rather than something completely informed. It does look like renderer code sometimes needs these magic numbers. If nothing else you did document your adjustment properly.

> because of the way I named this variable

That's possible too. It sounds like "shift zero" rather than some sort of tolerance or bias.

Kasasagi77:
0.000001 also does work - it just does not prevent all of the artifacts in my test cases. Even 0.000002 probably globally prevents like 99.999% of the artifacts or something.

You could make this value dependent on the texture size and what not and it would look better in the code than a magic number, but there is no need to complicate things like this IMO and it would only obscure the true nature of this fix.

PR #3644, opened 2026-05-23
feat: add background loading for boot, versus screen, and Turns

K4thos:
Add asynchronous loading paths for both boot-time select assets and pre-match fight assets, improving boot loading time and removing visible black-screen stalls. The default config keeps the previous behavior: match background loading, Turns preloading, and immediate boot/select background loading are all disabled by default.

Feats:

* Add `Config.VsScreenLoading` to start match asset loading during the versus screen instead of after it.
* Feed characters into the loader as their order slots are confirmed.
* Add versus-screen loading and ready motif elements, plus fallback wait UI/storyboard support when the versus screen is disabled.
* Add `waitforloading` storyboard support so loading storyboards can hold on the final scene until local loading and netplay readiness are complete.
* Add loader cancellation and cleanup for partially loaded match assets.
* Add a netplay loading handshake so both peers leave loading together.
* Add `Config.TurnsLoading` to preload the next Turns member during the match instead of between the matches.
* Add `Config.BootLoadingMode` for boot/select asset preloading:
  * `0`: existing behavior, preload before menus/select.
  * `1`: preload in background but wait before entering select.
  * `2`: enter menus/select immediately and materialize portraits/previews as they become ready.

### DIFF
--- a/external/script/main.lua
+++ b/external/script/main.lua
@@ -1542,7 +1542,7 @@ end
 function main.f_storyboard(path)
 	local s = loadStoryboard(path)
 	if s == nil then
-		return
+		return false
 	end
 	if gameOption('Debug.DumpLuaTables') then
 		-- get filename without extension from full path
@@ -1556,6 +1556,7 @@ function main.f_storyboard(path)
 		end
 		refresh()
 	end
+	return storyboardCanceled()
 end
 
 function main.f_hiscore(mode, place)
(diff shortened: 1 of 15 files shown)


K4thos:
under discussion, starting from here: https://discord.com/channels/233345562261323776/233363722934943744/1507791842199081212

TheRonix09:
I hope this is added

PR #3666, opened 2026-06-03
feat: Add DistOffset sctrl

Alceris:
I got sick of unwanted side switching mid-combo for various reasons or just when a character moves a large distance in one frame, so I made a simple and easy fix.

DistOffset: Applies an offset to the character's position for determining p2dist, autoturn, and which direction characters push each other.

It must apply to p2dist due to the prevalence of 'p2dist X < 0' being used as a trigger for the Turn sctrl.

Syntax is the same as Offset, with the addition of the Z axis. Comes with a DistOffset X/Y/Z trigger as well.

### DIFF
--- a/src/bytecode.go
+++ b/src/bytecode.go
@@ -980,6 +980,9 @@ const (
 	OC_ex3_analog_righty
 	OC_ex3_analog_lefttrigger
 	OC_ex3_analog_righttrigger
+	OC_ex3_distoffset_x
+	OC_ex3_distoffset_y
+	OC_ex3_distoffset_z
 	OC_ex3_helpervar_clsnproxy
 	OC_ex3_helpervar_helpertype
 	OC_ex3_helpervar_id
@@ -4195,6 +4198,13 @@ func (be BytecodeExp) run_ex3(c *Char, i *int, oc *Char) {
 		sys.bcStack.PushF(c.analogAxes[4])
 	case OC_ex3_analog_righttrigger:
 		sys.bcStack.PushF(c.analogAxes[5])
+	// DistOffset
+	case OC_ex3_distoffset_x:
+		sys.bcStack.PushF(c.distOffset[0] / oc.localscl) // Already in c.localscl so we only divide by oc.localscl
+	case OC_ex3_distoffset_y:
+		sys.bcStack.PushF(c.distOffset[1] / oc.localscl)
+	case OC_ex3_distoffset_z:
+		sys.bcStack.PushF(c.distOffset[2] / oc.localscl)
 	// HitDefVar
 	case OC_ex3_hitdefvar_guard_dist_width_back:
 		sys.bcStack.PushF(c.hitdef.guard_dist_x[1] * (c.localscl / oc.localscl))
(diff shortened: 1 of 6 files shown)


K4thos:
I've never heard of side switching mid-combo being something that can't be resolved cleanly in CNS code. This kind of solution looks like a workaround to me, but maybe I'm missing something since I don't have experience with char creation. What's the usual way to handle this behavior in char code? Let's see if it's hacky enough to warrant an sctrl just for this.

potsmugen:
The "large distance" thing has some legitimacy, but I think this solution is as hacky as any CNS code. It's quite literally fooling the engine. Having a hack done natively doesn't make it a good one.

Alceris:
Currently, there is no possible way to handle this in character code because there's no trigger to determine if playerpush is active or not. If there was, then you'd have to have a RunLast helper that checks if the characters are colliding and if they're not too far past each other, then move at least one in front of the other. By that point, the characters may have already turned if they were going to turn to face the opponent.

Alceris:
Additionally, by this same logic that it's 'fooling the engine,' and that that's bad for some reason, GroundLevelOffset also shouldn't exist. Both of these let us do things that are not otherwise possible without much more severe jank.

Alceris:
I just checked and GroundLevelOffset comes from you, PotS, so you're just actually being hypocritical.

K4thos:
## Merge decision vote

This vote is for active developers who have contributed to this project in the last 12 months.

Voting closes: 2026-07-04 23:59 UTC.

Please vote with one reaction:

- 👍 Merge
- 👎 Do not merge

Only votes from eligible active developers will be counted. Community feedback is still welcome.

Windblade-GR01:
I fell that having `turnOffset` would be better
As would affect only the character that called it.
.
Does the same but is way way less destructive.
This is bound to mess up interactions based on P2Dist the enemy characters use.
A lot of them begin graphical, so chaos would ensue.

Also, you would need to edit the target character anyway to implement this.
If you are editing the character anyway you can also change `p2dist X < 0` triggers.
You can literally do this in seconds in VSCode with the replace tool.
I'm really against having such destructive trigger were the only benefit over `turnOffset` would be to avoid the creator having to waste a minute or 2 to edit P2Dist.

The idea of a trigger that affects turn distance is not bad tho.
On my end I consider we should maybe go the the planning board and plan how to implement this in a less destructive way.

Alceris:
> I fell that having `turnOffset` would be better As would affect only the character that called it. . Does the same but is way way less destructive. This is bound to mess up interactions based on P2Dist the enemy characters use. A lot of them begin graphical, so chaos would ensue.
> 
> Also, you would need to edit the target character anyway to implement this. If you are editing the character anyway you can also change `p2dist X < 0` triggers. You can literally do this in seconds in VSCode with the replace tool. I'm really against having such destructive trigger were the only benefit over `turnOffset` would be to avoid the creator having to waste a minute or 2 to edit P2Dist.
> 
> The idea of a trigger that affects turn distance is not bad tho. On my end I consider we should maybe go the the planning board and plan how to implement this in a less destructive way.

Please explain why you would have to edit the target character for this. Not having to do that is, like, half the point of this. The primary feature is offsetting the center point of the character for which direction characters get pushed when their push boxes intersect, offsetting p2dist is just a necessary part of that to prevent characters from turning the wrong direction.

K4thos:
As per the general consensus among the devs, we’ve decided not to move forward with this change for now, so I’m closing this PR.

PR #3700, opened 2026-06-12
feat(lua): fade userdata API

wily-coyote:
This PR adds several new Lua functions to create and manage Fade userdata objects which are quite common in motif Lua. These objects are also used in the `fadeInInit`/`fadeOutInit` functions.

For module developers, the new functions are used in the same style as the Anim and TextImg APIs. Following is a list of the added functions:

- `fadeNew()`: Creates a new Fade userdata object with default values, or with a table with the following keys:
  - `time`: Length in ticks
  - `color`: R-G-B values in a Lua table. Range 0-255
  - `anim`: Anim userdata
  - `sound`: Group-index values in a Lua table. Sound from motif.

### DIFF
--- a/src/script.go
+++ b/src/script.go
@@ -2636,9 +2636,9 @@ func systemScriptInit(l *lua.LState) {
 		if !ok {
 			userDataError(l, 1, f)
 		}
-		f.col[0] = int32(numArg(l, 2))
-		f.col[1] = int32(numArg(l, 3))
-		f.col[2] = int32(numArg(l, 4))
+		f.col[0] = int32(numArg(l, 2)) & 0xff
+		f.col[1] = int32(numArg(l, 3)) & 0xff
+		f.col[2] = int32(numArg(l, 4)) & 0xff
 		return 0
 	})
 	luaRegister(l, "fadeSetAnim", func(*lua.LState) int {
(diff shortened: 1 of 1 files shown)


K4thos:
fadeSetFadeIn sounds weird to me. fadeTypeToggle would be more self-explanatory, I think. But do we even need it? It wasn't needed anywhere in the base scripts.
fadeNew is useful for modules (forgot to bring it back), but I don't see a use case for all the other set functions. If you don't have actual module in mind that would require them to work, I think the parameters could be sent as an optional table when calling fadeNew - see how it's done in animReset, for example.

wily-coyote:
I decided to remove `fadeSetFadeIn` since it's being handled by `fadeInInit` and `fadeOutInit` anyway.
I also added (optional) table-based initialization to `fadeNew` as requested.
However I decided to keep individual Lua functions since other userdata implementations also have them separated in the same way and I feel like it would make sense if fade also had it for consistency

K4thos:
Thanks but regarding the consistency - if there will be never a module that takes advantage of this, adding 4 new functions just for consistency doesn't sound like a good idea. It's just a bloat and more docs for users to comprehend, imo

K4thos:
Closed by mistake, the init function definitely will be merged, others still under discussion (looking for a use case)

wily-coyote:
I'll keep the individual functions off the PR, if an use case comes up in the future somebody else can come along and open a separate one for them. Should be OK now

PR #3770, opened 2026-07-08
feat: Loading additional textures for custom shader

NeatUnsou:
Shader parameters
・float sTime;
This value represents the number of frames since the shader was called. The count stops when the pause process is entered.

・uniform sampler2D tex1;
・uniform sampler2D tex2;
Can store additional textures loaded from the Anim number or SFF sprite number. Note that since textures are loaded as raw data, images with palettes may not display correctly as is.

ShaderSet/Explod/Projectile Parameters

・ShaderTex1.spr = Group, No
・ShaderTex1.anim = AnimNo

・ShaderTex2.spr = Group, No
・ShaderTex2.anim = AnimNo

Each of these will load the Anim or sprite specified in either of these as a texture. FightFX is not yet supported.

Explod/Projectile Parameters

・ShaderTime = int
Specifies the time the shader is active in Explod and Projectile. Default is -1 (persistent). In ShaderSet SCTRL, this is the Time parameter.

[kfm_zss_s.zip](https://github.com/user-attachments/files/29797730/kfm_zss_s.zip)
Ikemen_GO011

### DIFF
--- a/src/anim.go
+++ b/src/anim.go
@@ -903,7 +903,7 @@ func (a *Animation) drawSub1(angle, facing float32) (h, v, agl float32) {
 func (a *Animation) Draw(window *[4]int32, x, y, xcs, ycs, xs, xbs, ys,
 	rxadd float32, rot Rotation, rcx float32, pfx *PalFX, facing float32,
 	airOffsetFix [2]float32, projectionMode int32, fLength float32, color uint32,
-	isReflection bool, shader string, shaderParams [16]float32) {
+	isReflection bool, shader CustomShaderRenderData) {
 
 	// Skip blank animations
 	if a == nil || a.isBlank() {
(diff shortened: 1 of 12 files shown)


ToukaChan5900:
Cool!
but IMO, it makes more sense for shadertime to default to -1, instead of 0. So it'd make the parameter complettely optional.

NeatUnsou:
> Cool! but IMO, it makes more sense for shadertime to default to -1, instead of 0. So it'd make the parameter complettely optional.

It's true that a default of -1 is easier to use.
Please wait a moment.

ToukaChan5900:
The implementation looks very easy to understand and use. Pretty cool indeed! i just hope that the texture loading doesn't impact much the performance

gerarreal:
This is an incredible addition. Thank you Neat.

K4thos:
I'd like to merge this right away, but I noticed the new shader features aren't documented on the wiki yet. Are you planning to add docs there? We're preparing for the 1.0 release soon, but before that happens all features should be properly documented.

NeatUnsou:
I don't plan to make any major changes to the current specifications from here on out, so I think I can write some simple documentation.

potsmugen:
> > Cool! but IMO, it makes more sense for shadertime to default to -1, instead of 0. So it'd make the parameter complettely optional.
> 
> It's true that a default of -1 is easier to use. Please wait a moment.

I think your original idea was better since it was like PalFX. Now explod shaders default to off but with infinite time, which is a bit unusual. But it's just a detail.

PR #3780, opened 2026-07-14
fix: prevent stage select confirming on the character select press

mathmati:
This is a bug I encountered on the browser version and the cause is in the main branch too, not sure its encountered in the main game version.

In f_selectScreen(), the confirm press that completes character/palette selection is still readable by getInput() later in the same frame's pass, so the stage bar can confirm itself instantly on slot 0 (random). The stage that loads then mismatches the label the player last saw.

Ignore the confirm on the first frame the stage menu is active, that frame carries the leftover selection press; a fresh press is required to choose the stage.

### DIFF
--- a/external/script/start.lua
+++ b/external/script/start.lua
@@ -13,6 +13,7 @@ start.challenger = 0
 local restoreCursor = false
 local selScreenEnd = false
 local stageEnd = false
+local stageMenuReady = false
 local stageRandom = false
 local stageListNo = 0
 local t_aiRamp = {}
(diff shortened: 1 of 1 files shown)


K4thos:
I can't reproduce this issue on either the nightly or your WebAssembly build. Before the fix is merged, we need to verify that the issue actually exists, so reproduction steps or more details are needed. We discuss the reproducibility of it is here: https://discord.com/channels/233345562261323776/233363722934943744/1526607191266693211

Also, while AI assistance is allowed, generated comments should still be reviewed manually. Generally, the more comments the better, but four lines explaining something this simple feels excessive. I would either remove the comment entirely or shorten it to something like:
```lua
-- Ignore the carried-over confirm input to avoid instantly selecting a random stage.
```

K4thos:
Apparently the bug was found only in the browser version, so before merging it we might as well wait for it and test afterwards. btw. @mathmati feel free to open the pr for upstreaming the wasm build

PR #3801, opened 2026-07-24
feat: manual guarding

Alceris:
Guard sctrl: Has one parameter, GuardFlag, which determines which directions you are guarding. Syntax is the same as HitDef or HitOverride GuardFlag parameter.

AssertSpecial NoGuardState flag: Prevents character from automatically entering states 120, 150, 152, and 154.

### DIFF
--- a/src/bytecode.go
+++ b/src/bytecode.go
@@ -10370,6 +10370,29 @@ func (sc hitOverride) Run(c *Char, _ []int32) bool {
 	return false
 }
 
+type guard StateControllerBase
+
+const (
+	guard_guardflag byte = iota
+	guard_redirectid
+)
+
+func (sc guard) Run(c *Char, _ []int32) bool {
+	crun := getRedirectedChar(c, StateControllerBase(sc), guard_redirectid, "Guard")
+	if crun == nil {
+		return false
+	}
+
+	StateControllerBase(sc).run(c, func(paramID byte, exp []BytecodeExp) bool {
+		switch paramID {
+		case guard_guardflag:
+			crun.guardflag = exp[0].evalI(c)
+		}
+		return true
+	})
+	return false
+}
+
 type pause StateControllerBase
 
 const (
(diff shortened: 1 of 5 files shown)


gerarreal:
That's really awesome. One question, can you please add a check so attacks from behind can be unblockable?

Alceris:
> That's really awesome. One question, can you please add a check so attacks from behind can be unblockable?

Checking for if an attack is hitting from behind and then turning off guarding if so isn't hard to do in ZSS, so that seems like feature bloat to me.

Alceris:
> @Alceris this PR requires a wiki update before it can be merged.
> 
> Please:
> 
>     * update the relevant wiki page(s);
> 
>     * mark newly added features with **`(nightly build only)` directly in the relevant heading**;
> 
>     * reply here with the wiki page link(s) when finished.
> 
> 
> A maintainer will review the documentation and remove the blocking label when it is ready.

Done.
https://github.com/ikemen-engine/Ikemen-GO/wiki/State-controllers-(new)#guard-nightly-build-only
https://github.com/ikemen-engine/Ikemen-GO/wiki/State-controllers-(changed)#noguardstate-nightly-build-only

K4thos:
I can't reopen the merged PR, so feel free to open a new one. The proposed implementation is under discussion.
Context: https://discord.com/channels/233345562261323776/233363722934943744/1530013321569243197
I don’t have a strong opinion on how it should work, but we should reach some consensus among devs on the best approach.
I'm adding the docs here in case this gets merged later.

## <a name="new_guard">Guard (nightly build only)</a>  

Sets the player to be guarding in the specified directions, regardless of their current condition.

**Required parameters:**  
  
>guardflag = *hit_flags* (string)  
The directions to guard, generally H, L, M, and/or A.

**Example:**
```ini
[State Test]
type = Guard
trigger1 = Command = "holdz"
guardflag = MA
```


#### <a name="changed_assertspecial_flag_noguardstate">NoGuardState (nightly build only)</a>

Disables the hard-coded state transitions to States 120, 140, 150, 152, and 154.

PR #3830, opened 2026-07-28
feat: Add debug dump feature via stdout

NeatUnsou:
・Added the -debugdump command line flag (requires AllowDebugMode). runMatch() loop to call the global Lua function __ikemen_debug_dump and print its result as JSON to stdout.

・Added several getter/setter Lua functions to allow dynamic reading and writing of properties.

・Added a breakpoint check system that evaluates Lua code strings every frame and triggers sys.paused = true if the condition is met.

・Added breakpoint management functions.

-debugdump command line flag and extends Lua bindings to facilitate the creation of external debugging tools. This approach allows the engine to evaluate a user defined Lua function (__ikemen_debug_dump) at the each frame and output the returned table as a JSON string to stdout. External tools can interact with the engine by reading this output and sending Lua commands via stdin.

Example Use Case (External Tool):
Since the data extraction logic is entirely driven by Lua scripts sent through stdin, external GUI tools can be developed independently in any language. 
For example, a Python script can run IKEMEN GO via subprocess, inject the __ikemen_debug_dump function to request specific data (vars, explods, projectiles), and read the JSON output.
[debugdump_gui.py](https://github.com/user-attachments/files/30481164/debugdump_gui.py)

### DIFF
--- a/src/main.go
+++ b/src/main.go
@@ -180,6 +180,16 @@ func realMain() {
 	sys.cfg = *cfg
 	// Logcat("LOG: Config Loaded. System Script: " + sys.cfg.Config.System)
 
+	if _, ok := sys.cmdFlags["-debugdump"]; ok {
+		if sys.cfg.Debug.AllowDebugMode {
+			sys.debugDumpEnabled = true
+			sys.breakpoints = make(map[int]*Breakpoint)
+			Logcat("Debug dump mode enabled.\n")
+		} else {
+			Logcat("Debug dump not started because AllowDebugMode is false\n")
+		}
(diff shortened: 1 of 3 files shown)


K4thos:
As per the general consensus among the devs, we've decided not to move forward with this change in its current form, so I'm closing this PR.

A smaller version that provides a debug dump for external tools to consume and work with would likely be fine, so feel free to revisit that approach in a new PR. The other part, however, would introduce additional maintenance on the engine side, which we don't think is the right direction for this.

There's some more context on Discord.

PR #3988, opened 2026-09-11
fix: guard and recovery frame timing

### LINKED ISSUE #3916: Inconsistency between the frame where blocking becomes possible and the frame where state transitions become possible
### Describe the bug

mmediately after a state transition, on the exact frame where ctrl becomes active (ctrl = 1), normal state transitions (such as executing moves that require ctrl = 1) cannot be performed yet, but blocking is somehow possible on this exact frame.
In MUGEN, the frame where blocking becomes possible is identical to the frame where state transitions become possible, so this issue does not occur.

### To Reproduce

[test_kfm.zip](https://github.com/user-attachments/files/31432489/test_kfm.zip) in both IKEMEN GO and MUGEN 1.1.

1. Select the same character for both P1 and P2.
2. Stand close to each other.
3. Have P1 press the 'b' button to attack.
4. Have P2 walk backward to attempt blocking.
5. In MUGEN 1.1, P2 fails to block the attack, but in IKEMEN GO, P2 successfully blocks it.

### Expected behavior

The block should fail, exactly as it does in MUGEN 1.1.

### Screenshots / Video

MUGEN 1.1

https://github.com/user-attachments/assets/2277c694-b19e-437d-b23b-2b02cfc50428

IKEMEN GO

https://github.com/user-attachments/assets/766c029c-ab2f-4045-a729-829b06256148

v1.0 rc4

### Operating system

Windows

I suspect this current behavior in IKEMEN GO is a side effect of a workaround designed to fix a well-known MUGEN bug where the very first frame after recovering from a gethit state is unguardable.
Reference: [https://w.atwiki.jp/mugenarcihve/pages/48.html](https://www.google.com/url?sa=E&q=https%3A%2F%2Fw.atwiki.jp%2Fmugenarcihve%2Fpages%2F48.html)
(In IKEMEN GO, the duration of the gethit state is extended by 1 frame compared to MUGEN, and in exchange, blocking is allowed on the 1st frame of recovery.)

While this intentional change successfully mitigates the unguardable frame issue from MUGEN regarding gethit recovery, it inadvertently affects all state transitions where ctrl switches from 0 to 1. As a result, the frame advantage/disadvantage (plus/minus frames) of attacks is shifted by 1 frame compared to MUGEN.

NeatUnsou:
Fixes #3916

・Prevent Ctrl gained during the current state from enabling guard on the same frame. 
・Keep MoveType = H until the next tick when recovering after negative states have already run. 
・Remove the legacy hittime + 1 compensation, which is no longer needed.

Guard evaluation still happens after state processing, so the MUGEN behavior where a character can guard on the first frame of an attack remains fixed.

### DIFF
--- a/src/bytecode.go
+++ b/src/bytecode.go
@@ -10653,7 +10653,7 @@ func (sc stateTypeSet) Run(c *Char, _ []int32) bool {
 		case stateTypeSet_statetype:
 			crun.ss.changeStateType(StateType(exp[0].evalI(c)))
 		case stateTypeSet_movetype:
-			crun.changeMoveType(MoveType(exp[0].evalI(c)), true)
+			crun.ss.changeMoveType(MoveType(exp[0].evalI(c)))
 		case stateTypeSet_physics:
 			crun.ss.physics = StateType(exp[0].evalI(c))
 		}
--- a/src/bytecode.go
+++ b/src/bytecode.go
@@ -10653,7 +10653,7 @@ func (sc stateTypeSet) Run(c *Char, _ []int32) bool {
 		case stateTypeSet_statetype:
 			crun.ss.changeStateType(StateType(exp[0].evalI(c)))
 		case stateTypeSet_movetype:
-			crun.ss.changeMoveType(MoveType(exp[0].evalI(c)))
+			crun.changeMoveType(MoveType(exp[0].evalI(c)), true)
 		case stateTypeSet_physics:
 			crun.ss.physics = StateType(exp[0].evalI(c))
 		}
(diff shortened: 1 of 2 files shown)


NeatUnsou:
This fix attempts to mitigate the MUGEN specification that prevents operation during recovery frames by treating recovery frames as MoveType=H.

This will continue to resolve issues such as damage scaling being interrupted, target loss, and nochainID loss during recovery frames.

Furthermore, since the guarding timing will be the same as in MUGEN, issue #3916 will also be resolved.

K4thos:
As per the general consensus among the devs, we’ve decided not to move forward with this change for now, so I’m closing this PR.

PR #4008, opened 2026-09-15
fix: savestate aliasing; refactor: rollback logging

potsmugen:
Fixes:
- Delete a savestate entry when returning it to the pool, so two frames can't end up aliasing the same state

Refactor:
- Increase number of logged rollback states to capture the worst case scenario
- Added the random seed to rollback state logs
- Split the options for saving the full GGPO logs and the state logs

### DIFF
--- a/.github/workflows/wiki-docs-gate.yml
+++ b/.github/workflows/wiki-docs-gate.yml
@@ -1,0 +1,32 @@
+name: Wiki documentation gate
+
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - labeled
+      - unlabeled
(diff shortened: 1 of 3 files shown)


K4thos:
PotS, your repo picked up my old 3e81055 build commit, so merging it as-is would re-add the deleted wiki-docs-gate.yml/wiki-docs-notify.yml. Mind if I force-push a version rebased on current develop with just your three commits?

potsmugen:
Yeah I'm not sure why it did that. I think I need to re-add the dev branch. But sure, go ahead. I can also make a fixed PR if you want.

K4thos:
yeah, please make a new pr if it's not a problem for you

PR #4022, opened 2026-09-17
Adjust combo logic for hits and trueHits conditions (made it that combos are already shown on the very first hit)

menyek67:
Made that combos are already shown on the very first hit for damage debugging, or it would be great if there would be a setting starting at what hit the combo counter plus the damage in percentage, for damage debugging.


K4thos:
changing whole engine behavior just so you can debug better doesn't make sense. The way it works is intentional. You have triggers and DisplayToClipboard/PrintToConsole/Text state controllers to print this data anyway.

